PR #12, opened 2015-09-03
Add ros_tools utility package

jpanikulam:
Includes simple helpers for rospy:
- Image handling
- Arbitrary message unpacking
- wait_for_param
- thread locking help


utils/sub8_ros_tools/sub8_ros_tools/image_helpers.py
```diff
@@ -0,0 +1,58 @@
...
+        self.im_pub = rospy.Publisher(topic, Image, queue_size=queue_size)
+        self.bridge = CvBridge()    
+        self.encoding = encoding
+    
+    def publish(self, cv_image):
+        try:
+            image_message = self.bridge.cv2_to_imgmsg(cv_image, self.encoding)
+            self.im_pub.publish(image_message)
+        except CvBridgeError, e:
+            print e
```
pemami4911: Use ROS_ERROR here?

utils/sub8_ros_tools/sub8_ros_tools/image_helpers.py
```diff
@@ -0,0 +1,58 @@
...
+        self.im_sub = rospy.Subscriber(topic, Image, self.convert, queue_size=queue_size)
+        self.bridge = CvBridge()
+        self.callback = callback
+
+    def convert(self, data):
+        try:
+            image = self.bridge.imgmsg_to_cv2(data, desired_encoding=self.encoding)
+            self.callback(image)
+        except CvBridgeError, e:
+            print e
```
pemami4911: Same ^

utils/sub8_ros_tools/sub8_ros_tools/math_helpers.py
```diff
@@ -0,0 +1,25 @@
+import numpy as np
+
+
+def rosmsg_to_numpy(rosmsg, keys=None):
```
pemami4911, line 4: Might want to be more specific with the function name, since "rosmsg" might suggest you can pass in a msg that is not a ROS vector or a quaternion

utils/sub8_ros_tools/sub8_ros_tools/math_helpers.py
```diff
@@ -0,0 +1,25 @@
...
+    '''
+    if keys is None:
+        keys = ['x', 'y', 'z', 'w']
+    output_array = []
+    for key in keys:
+        # This is not necessarily the fastest way to do this
+        if hasattr(rosmsg, key):
+            output_array.append(getattr(rosmsg, key))
+        else:
+            break
```
pemami4911: Throw a non-fatal exception here (read: "raise") in case the rosmsg doesn't have [x, y, z, w], which will make it easy for clients of this utility function to know that the conversion failed

utils/sub8_ros_tools/test_ros_tools/test_ros_tools.py
```diff
@@ -0,0 +1,48 @@
...
+        '''Test that a made ros image can be returned to its original form'''
+        im = np.ones((255, 255, 3)).astype(np.uint8)
+        im_msg = make_image_msg(im)
+
+        cv_im = get_image_msg(im_msg)
+        np.testing.assert_array_equal(im, cv_im)
+
+
+if __name__ == '__main__':
+    unittest.main()
```
pemami4911, line 58: Add "sad path" test cases for the utility functions, like passing in an empty rosmsg to "rosmsg_to_numpy", to verify it won't cause a crash!

forrestv:
FWIW, trying to deal with functions being handed improper types is generally a lost cause. If a function is given a parameter with a wrong type, it's a programmer error, and there's really no good way to deal with it. You should just abort immediately/crash, either by just letting something weird happen (e.g. trying to access `.x` of an object without an x attribute, which will trigger an `AttributeError`, which shouldn't be caught) or having an assertion fail (e.g. `assert isinstance(rosmsg, Quaternion)`). The first one is actually probably preferable to the second to go along with Python's duck typing -- you shouldn't care what type something is, just if it can do what you want.

Read https://giantfublog.wordpress.com/2015/03/11/errors-and-exceptions/ . This is what they call a "hard error" (since it's a violation of precondition).

pemami4911:
Fair enough. In this case then, raising an exception when a programmer mis-uses "rosmsg_to_numpy" wouldn't be necessary. It would be helpful, though, to be more explicit with the name of the function- to help programmers use it correctly!

zachgoins:
It would be helpful to include the features of 'orientation_helpers.py' in
the 'math_helpers.py' file. Allow users to not only convert from
ROSmsgs/Quaternions but then to go between quaternions and rotation
vectors, normalize, get perpendicular/orthogonal vectors, find determinant,
find inverse, verify inverse with identity matrix, etc...

https://github.com/uf-mil/software-common/blob/master/uf_common/src/uf_common/orientation_helpers.py

jpanikulam:
@pemami4911, Changes made, up for review again in the same PR

utils/sub8_ros_tools/sub8_ros_tools/math_helpers.py
jpanikulam, line 4: Instead of renaming the function, made it work for an arbitrary message

jpanikulam:
@forrestv Made the thread_lock change as well

Issue #4, opened 2015-08-27
Update simulation


jpanikulam:

jpanikulam:
Now satisfied by Gazebo sim.

PR #17, opened 2015-09-21
SIM: Implement simulation

jpanikulam:
See [Simulation wiki page](https://github.com/uf-mil/Sub8/wiki/Simulation) for considerations,

and [Simul8 wiki page](https://github.com/uf-mil/Sub8/wiki/Simul8) for what is in, and what is not.

The sim is now functional enough to test controllers and mappers, and therefore satisfies the first simulation design milestone. Refs #4 

Since this is already nearly 1,000 lines I'm submitting a PR now to receive early criticism.


pemami4911:
@zachgoins adding you to review this PR as well since it's a big one

pemami4911:
Would be helpful to reviewers if a quick explanation of how this code was tested/verified was included in the commit message

simulation/sub8_simulation/launch/sim_thrusters.launch
```diff
@@ -0,0 +1,67 @@
+<launch>
```
pemami4911: This looks like the same thing we have over in the sub8_thruster_mapper. Is there a way to prevent duplicating code?

simulation/sub8_simulation/sub8_sim_tools/physics/physics.py
```diff
@@ -0,0 +1,221 @@
...
+        # Create contact joints
+        for c in contacts:
+            c.setBounce(.4)
+            c.setMu(5000)
+            j = ode.ContactJoint(world, contact_group, c)
+            j.attach(geom1.getBody(), geom2.getBody())
+
+
+class Entity(object):
+    _linear_damping_coeff = -0.2
```
pemami4911: Do we want to be able to change these on a case by case basis,or is the desired functionality for these values to remain static?

jpanikulam:
Addressing comments:
- Only some sim components will be configurable, unless we decide otherwise
- Cutting down on repeated code in the physics components is difficult, because ODE is picky about how geometries and bodies are defined
- I eliminated repeated code in the launch files

zachgoins:
I have looked through all the files and cannot see any glaring problems just by looking. It would be very beneficial though to include a few lines at the top of each file to give an idea of what the goal of the file is in the grand scheme of the sim.

zachgoins:
Is is possible to run instances of the sim without rendering graphics? We could use that feature to run continuous integration tests in the background with the physics alone.

jpanikulam:
That is a good point. 

Right now, physics is completely decoupled from rendering, and this is a good argument to keep it that way. 

@annhiluc thoughts?
- Jake

jpanikulam:
I suppose we'd also want to simulate vision in that case. We may have to fork Vispy to make it possible. 
- Jake

Annhiluc:
Since physics is independent of rendering, we could make an instance without visualization. What would need to be displayed, if anything?

fnivek:
Anyone know about the Model View Controller design pattern? This seems like
a perfect place to use the MVC pattern.

On Tue, Sep 22, 2015 at 12:53 PM, Annie Luc notifications@github.com
wrote:

> Since physics is independent of rendering, we could make an instance
> without visualization. What would need to be displayed, if anything?
> 
> —
> Reply to this email directly or view it on GitHub
> https://github.com/uf-mil/Sub8/pull/17#issuecomment-142347046.

## 

Sincerely,
Kevin French,
University of Florida,
4th year Electrical Engineer,
Cell: Available upon request,
Email: fnivek@gmail.com

jpanikulam:
Reassigning to Patrick for final pull

PR #20, opened 2015-09-23
TGEN: Added implementation of OMPL framework

pemami4911:
Sub8StateSpace is an extension of OMPL's CompoundStateSpace.
The sub's state space is defined as a 12D vector. Currently,
the time derivative of the euler angles is used to be
the representation of the angular velocities,
and a unit quaternion is used for the orientation.

Planning will be carried out in the sub's control space;
accordingly, Sub8SpaceInformation is an extension of OMPL's
control::SpaceInformation.

StateValidityChecker allows the client to customize
collision checking, clearance from nearest invalid states,
etc.

Most unit tests are currently failing, since tests have
been written for unimplemented functions for Sub specific
behavior.

catkin_make was verified to build successfully.

### DIFF
--- a/gnc/sub8_trajectory_generator/include/sub8_trajectory_generator/sub8_state_space.h
+++ b/gnc/sub8_trajectory_generator/include/sub8_trajectory_generator/sub8_state_space.h
@@ -1,0 +1,182 @@
+/**
+ * Author: Patrick Emami
+ * Date: 9/21/15
+ */
+#ifndef SUB8_STATE_SPACE_H_
+#define SUB8_STATE_SPACE_H_
+
+#include "ompl/base/StateSpace.h"
+#include "ompl/base/spaces/RealVectorStateSpace.h"
+#include "ompl/base/spaces/SO3StateSpace.h"
+
+using ompl::base::CompoundStateSpace; 
+using ompl::base::RealVectorStateSpace; 
+using ompl::base::RealVectorBounds;
+using ompl::base::SO3StateSpace; 
+using ompl::base::StateSpacePtr;
+using ompl::base::State; 
+
+namespace sub8 {
+
+	namespace trajectory_generator {
+
+		// Defines a custom state space, derived from 
+		// ompl::base::CompoundStateSpace. 
+		//
+		// This state space consists of [q, qdot], where 
+		// [q] is defined as position and orientation in the world coordinate frame
+		// and [qdot] is the time rate of change of [q]. 
+		// The dynamics are modeled with xdot = Ax + Bu, where x = [q, qdot]
+		// and xdot = [qdot, qdoubledot]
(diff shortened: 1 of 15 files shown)


gnc/sub8_trajectory_generator/src/sub8_state_space.cc
```diff
@@ -0,0 +1,34 @@
+/**
+ * Author: Patrick Emami
+ * Date: 9/21/15
+ */
+
+#include "sub8_state_space.h"
+
+ompl::base::State* sub8::trajectory_generator::Sub8StateSpace::allocState() const {
+	StateType *state = new StateType();
```
jpanikulam, line 9: shared_ptr possible?

gnc/sub8_trajectory_generator/include/sub8_trajectory_generator/sub8_state_space.h
```diff
@@ -0,0 +1,182 @@
...
+
+			// Override the default distance metric, which is to use the L2 norm
+			// Instead, LQR  will be used to calculate a cost in the control space
+			// between two states
+			virtual double distance(const State* state1, const State* state2 ) const; 
+
+			// Get the maximum value a call to distance() can return (or an upper bound).
+			// For unbounded state spaces, this function can return infinity.
+			//
+    		// Tight upper bounds are preferred because the value of the extent is used
```
jpanikulam: Sudden tab change?

gnc/sub8_trajectory_generator/src/sub8_state_space.cc
forrestv, line 9: Why are you using time derivatives of Euler angles for angular velocity
instead of just the angular velocity vector? From your comments, it looks
like that's what you're actually doing already ("Get the angular velocity
about the y axis").

At least from what I've seen, all that does is introduce more complex
Jacobians everywhere and add singularities. It was only used in the old
trajectory generator because it worked directly on each Euler angle
separately, but AFAIK your approach won't.

pemami4911, line 9: Right, so this was still a point of confusion for me at the time. I'll just be using the angular velocity vector, I see now how using the time derivative of the Euler angles wouldn't go over very well in this situation. I'll make the relevant changes to reflect this.

pemami4911:
- The virtual function I'm overriding has a return value of ompl::base::State*, so using a raw pointer there is required and OK
- Changed a few function names in Sub8StateSpace to make the angular velocity getters more accurate, and fixed the tabulation problem (Sublime wasn't showing that there was spacing issues, but gedit did ?)

jpanikulam:
Re: spacing
Is it from mixing tabs and spaces?
- Jake

pemami4911:
Yep haha

jpanikulam:
It is done

PR #21, opened 2015-10-04
Patrick

pemami4911:
Large PR, but want to get this approved since it will just get larger if I don't push incrementally :cactus: 
Most of the changes to the unit tests are just clang auto-formatting. The only new test is for TGenManager.

See commit messages for more detail about the actual changes, but in summary: 
Added Sub8TGenManager and the Waypoint, Trajectory, and MotionPlan msgs/srvs. 
Began implementing the ODE governing the sub dynamics. 

Current planning is implemented with RRT and PDST - the planner can be selected by passing a param in the launch file. The planner still needs a distance function and collision checking and boundaries for the different subspaces, so it is non-functional. Current plan is to implement RRT\* for kinodynamic planning ourselves and to integrate that into this design, since RRT and PDST will most likely not be sufficient due to the fact that the quality of trajectories that they generate is usually quite poor.

### DIFF
--- a/gnc/sub8_trajectory_generator/include/sub8_trajectory_generator/sub8_state_space.h
+++ b/gnc/sub8_trajectory_generator/include/sub8_trajectory_generator/sub8_state_space.h
@@ -82,12 +82,12 @@ namespace sub8 {
 					return as<RealVectorStateSpace::StateType>(2)->values[2]; 
 				}
 
-				// Get a const reference to the orientation as a unit quaternion
+				// Get a const reference to the orientation (internally a unit quaternion)
 				const SO3StateSpace::StateType& rotation() const {
 					return *as<SO3StateSpace::StateType>(3);
 				}
 
-				// Get a reference to the orientation as a unit quaternion
+				// Get a reference to the orientation 
 				SO3StateSpace::StateType& rotation() {
 					return *as<SO3StateSpace::StateType>(3);
 				}
--- a/gnc/sub8_trajectory_generator/include/sub8_trajectory_generator/sub8_state_space.h
+++ b/gnc/sub8_trajectory_generator/include/sub8_trajectory_generator/sub8_state_space.h
@@ -34,12 +34,12 @@ class Sub8StateSpace : public CompoundStateSpace {
     // The as<SomeStateSpace::StateType>(i) call returns a reference
     // to the i'th subspace of the CompoundStateSpace, of type "SomeStateSpace"
 
-    // Get the x component of the position
+    // Set the x component of the position
     void setX(double x) const {
       as<RealVectorStateSpace::StateType>(0)->values[0] = x;
     }
 
-    // Get the y component of the position
+    // Set the y component of the position
     void setY(double y) const {
       as<RealVectorStateSpace::StateType>(0)->values[1] = y;
     }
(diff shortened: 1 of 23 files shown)


utils/sub8_msgs/msg/Waypoint.msg
```diff
@@ -0,0 +1,4 @@
+# Waypoint
+
+geometry_msgs/Pose pos
```
jpanikulam: Convention is to use pose and twist as the names of the members, see: [odometry msg definition](http://docs.ros.org/jade/api/nav_msgs/html/msg/Odometry.html)

i.e. 

```
geometry_msgs/Pose pose
geometry_msgs/Twist twist
```

gnc/sub8_trajectory_generator/src/trajectory_generator.cc
```diff
@@ -4,13 +4,68 @@
...
+    boost::shared_ptr<sub8_msgs::Waypoint> goal_state_wpoint =
+        boost::make_shared<sub8_msgs::Waypoint>(req.goal_state);
+
+    _tgen->setProblemDefinition(_tgen->waypointToState(start_state_wpoint),
+                                _tgen->waypointToState(goal_state_wpoint));
+
+    // What is the desired behavior here? If planning fails and no
+    // solution path is found, should we return false from the callback? 
+    // Or just set "resp.success" to false? If this happens, is it a 
+    // system-shutdown-worthy event?
```
jpanikulam, line 36: Do we have a good idea of when trajectory generation fails? Will it fail under predictable conditions? Will we always be stationary when planning?

gnc/sub8_trajectory_generator/src/sub8_tgen_manager.cc
```diff
@@ -0,0 +1,199 @@
...
+        static_cast<GoalState*>(pdef->getGoal().get())->getState());
+    solve();
+
+  } else {
+    ROS_INFO("%s", Sub8TGenMsgs::TRAJECTORY_VALIDATED);
+  }
+}
+
+State* Sub8TGenManager::waypointToState(
+    const boost::shared_ptr<sub8_msgs::Waypoint>& wpoint) {
```
jpanikulam, line 142: Clang-Format? Nice.

gnc/sub8_trajectory_generator/include/sub8_trajectory_generator/sub8_state_space.h
```diff
@@ -7,176 +7,218 @@
...
+    // pos[0] = x
+    // pos[1] = y
+    // pos[2] = z
+    void getPosition(std::vector<double>& pos) const {
+      if (!pos.empty()) {
+        pos.clear();
+      }
+      pos.push_back(as<RealVectorStateSpace::StateType>(0)->values[0]);
+      pos.push_back(as<RealVectorStateSpace::StateType>(0)->values[1]);
+      pos.push_back(as<RealVectorStateSpace::StateType>(0)->values[2]);
```
jpanikulam, line 60: Why std::vector? Does OMPL support Eigen?

jpanikulam:
I am getting a make error

```
[ 97%] Building CXX object Sub8/gnc/sub8_trajectory_generator/CMakeFiles/trajectory_generator.dir/src/sub8_state_validity_checker.cc.o
/home/jacob/repos/rosboat_ws/src/Sub8/gnc/sub8_trajectory_generator/src/sub8_space_information.cc:8:35: fatal error: sub8_state_propagator.h: No such file or directory
 #include "sub8_state_propagator.h"
                                   ^
compilation terminated.
```

gnc/sub8_trajectory_generator/include/sub8_trajectory_generator/sub8_state_space.h
pemami4911, line 60: Yeah, Eigen can be used here. I'm not too familiar with Eigen, and std::vector is perfectly adequate. But I imagine Eigen would make this code cleaner, so will look into refactoring.

jpanikulam:
Still alive?

gnc/sub8_trajectory_generator/src/trajectory_generator.cc
pemami4911, line 36: I would say that we can make the assumption that the start point for planning our trajectory will always be a "stationary" point. If we are in motion and we find out that our current trajectory has become invalid, I think for now the behaviour should just be to first issue a "stop" trajectory that attempts to halt the sub and keep it stationary, and then re-plan from the current position. 
As long as the assumption that valid start and goal states are passed to the TGEN, I don't think planning will ever fail - it may take a long time if the sub is surrounded by a large number of obstacles. So, a timeout error would be the only potential failure condition.

jpanikulam:
Merged.

PR #23, opened 2015-10-18
Implement Alarms

jpanikulam:
This is the first iteration of the alarm system.

Here is my [SemaphoreCI build](https://semaphoreci.com/jpanikulam/sub8/branches/alarms/builds/2) as proof that the build succeeds and all unit-tests pass.


jpanikulam:
[Some associated documentation.](https://github.com/uf-mil/Sub8/wiki/System-Alarms)

utils/sub8_msgs/msg/PoseTwist.msg
```diff
@@ -0,0 +1,3 @@
+std_msgs/Header header
```
pemami4911: This is the same as the Waypoint message I've already added. We only need to keep one or the other.

command/sub8_alarm/nodes/alarm_handler.py
```diff
@@ -0,0 +1,59 @@
...
+        '''Alarm Handler
+            Listen for alarms, call scenarios
+        TODO:
+            - Add alarm queue
+            - Handle alarms in sequence (Don't get stuck in the callback)
+            - bag (if set) EVERY single alarm received
+        '''
+        rospy.init_node('alarm_handler')
+        # Queue size is large because you bet your ass we are addressing every alarm
+        self.alarm_sub = rospy.Subscriber('/alarm', Alarm, self.alarm_callback, queue_size=100)
```
pemami4911, line 21: We should monitor this queue size value as the system progresses- it may prove to be too small if multiple things fail at once, and lost alarms would be sad

command/sub8_alarm/sub8_alarm/alarm_helpers.py
```diff
@@ -0,0 +1,79 @@
...
+
+        # Notify the world that our alarm has spawned
+        # For now, unused - future: Use for system reports
+        # self.alarm_spawn_pub = rospy.Publisher('/alarm_spawn', Alarm, latch=True, queue_size=1)
+
+        # Get ROS name
+        self.node_name = rospy.get_name()
+
+        # Alarms associated with this broadcaster
+        self.alarms = []
```
pemami4911, line 21: What's the benefit of storing your alarms in a list here?

utils/sub8_msgs/msg/PoseTwist.msg
jpanikulam: Oops - added this before the waypoint. Removing.

jpanikulam: In the future, the controller will _only_ take trajectory messages, so that we can set frames. It would be redundant to force each waypoint message to carry a full header.

command/sub8_alarm/nodes/alarm_handler.py
jpanikulam, line 21: Good point - adding to future notes

command/sub8_alarm/sub8_alarm/alarm_helpers.py
jpanikulam, line 21: Future plans to do some bookkeeping in the background. Will note in comments.

Also worth forcing the association of Alarms _belonging_ to their AlarmBroadcasters

pemami4911:
Looks goot

command/sub8_alarm/sub8_alarm/alarm_helpers.py
```diff
@@ -0,0 +1,80 @@
...
+        # self.alarm_spawn_pub = rospy.Publisher('/alarm_spawn', Alarm, latch=True, queue_size=1)
+
+        # Get ROS name
+        self.node_name = rospy.get_name()
+
+        # Alarms associated with this broadcaster
+        # In the future, there will be some background bookkeeping
+        self.alarms = []
+
+    def add_alarm(self, name, action_required=False, severity=5, problem_description=None, parameters=None):
```
pemami4911, line 23: Update wiki with severity levels; the wiki currently only has descriptions for levels 0-2

jpanikulam, line 23: Ohp, changing the code. I did this because I glanced at the wiki, and the
biggest number I saw was "5"

Jacob Panikulam | (630) - 877 - 7665
Software Lead, 2016 Subjugator Team http://subjugator.org/
Github https://github.com/jpanikulam | About Me
http://jpanikulam.github.io/

command/sub8_alarm/sub8_alarm/alarm_helpers.py
```diff
@@ -0,0 +1,80 @@
...
+        self._action_required = action_required
+        self._severity = severity
+        self._problem_description = problem_description
+        self._parameters = parameters
+
+        self._alarm_pub = alarm_publisher
+
+    def raise_alarm(self, problem_description=None, parameters=None):
+        '''Arguments are override parameters'''
+        assert (problem_description != "") or (self._problem_description is not None), "No problem description has been set for this alarm"
```
pemami4911, line 55: Do we want to leave this assert here later on? Good for debugging, but during production If an alarm doesn't have a problem description, isn't it better just to let it through (since the alarm will still go off and be handled)?

jpanikulam, line 55: Let's remove this in the next iteration - 

I think in Alarms 2.0 there should be all of the asserts in the alarm _declaration_ but no asserts in the alarm _raising_, because once an alarm is raised we should do as little as possible to trigger a potential error

jpanikulam, line 55: I have started a list of desirements for Alarms 2.0 at the bottom of [this page](https://github.com/uf-mil/Sub8/wiki/System-Alarms)

PR #24, opened 2015-10-28
Sim: Add features as prelude to monte-carlo tools

jpanikulam:
- Package structure for  simple controller
- Prototype class for simple controller
- Some message tools
- Allow the sim to run faster (Or slower, if you want...) than real-time
- Allow the sim to run without rendering anything (Pure physics)

[My Semaphore](https://semaphoreci.com/jpanikulam/sub8/branches/sim/builds/3)


simulation/sub8_simulation/nodes/simulate.py
```diff
@@ -55,21 +59,65 @@ def add_box(self, position, density, width, height, depth, color):
...
+            ("BRL", np.array([ 0.866, -0.5,  0]), np.array([-0.2678, -0.2795, 0.0000])),
+        ]
+
+        # add the thrust indicators
+        for name, rel_direction, rel_position in thruster_list:
+
+            R = make_rotation(np.array([0.0, 0.0, 1.0]), rel_direction)
+            transformation = compose_transformation(np.transpose(R), rel_position)
+
+            class ThrustGetter(object):
```
jpanikulam, line 91: This is crappy right now - in the next iteration, I will move this into a "widget" in a seperate file

PR #29, opened 2015-11-06
Monte-Carlo: Add simple controller verification

jpanikulam:
I am expecting the controller integration test to fail, because the controller is not "implemented" in the sense that it has not yet been PR'd. 
# What This Did
- Add verification tool
- Add examples
- Add tests
- Fixed a race condition that would rise _during_ thruster drop alarms
- Made alarm integration test behave more like the real sub


command/sub8_alarm/test_alarms/test_alarm_integration.py
```diff
@@ -46,15 +46,19 @@ def test_lose_thruster(self):
...
 
         time.sleep(0.1)
-        wrench_pub.publish(WrenchStamped(
-            # The actual force/torque is irrelevant, we still send a 0 command for every active thruster
-            wrench=Wrench(
-                force=Vector3(10.0, 10.0, 10.0),
-                torque=Vector3(10.0, 10.0, 10.0)
-            )
-        ))
+        give_up_time = time.time() + 1.5
```
pemami4911: A comment here explaining your reasoning behind this change would be helpful. Why did you opt to use 1.5 as your timeout? Is it just arbitrary?

simulation/sub8_montecarlo/montecarlo_tests/montecarlo_pd_controller.py
```diff
@@ -0,0 +1,52 @@
...
+    def setUp(self, *args):
+        '''TODO:
+            - Assert that wrenches within bounds are close to the unbounded least-squares estimate
+            - Make waiting for node functionality into a test_helpers util
+            - Test random wrenches and assert valid behavior
+        '''
+        self.num_runs = rospy.get_param('runs', 4)
+        self.time_limit = rospy.get_param('run_time', 10)
+
+    def test_pd_controller(self):
```
pemami4911: You'll probably want to rename this unit test at some point to make the name more accurately represent what you're testing here. This is especially the case if you're going to be adding more unit tests for the controller.

simulation/sub8_montecarlo/nodes/example_verification.py
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env python
+'''Monte-Carlo controller verification
+'''
+import rospy
+from sub8_montecarlo_tools import VerifyController
+
+
+if __name__ == '__main__':
+    rospy.init_node('verify_pd')
```
pemami4911: Does this really need to be in a separate file? It looks like this could be moved into a comment in controller_verify.py. This code is already being ran in your PD controller test as well; an option is to write a separate unit test for this basic use case of the VC class though.

simulation/sub8_montecarlo/montecarlo_tests/montecarlo_pd_controller.py
```diff
@@ -0,0 +1,52 @@
+#!/usr/bin/env python
+
```
pemami4911: I don't think we added a blurb about naming conventions for unit tests to the Code Policy, but I think up to now we've been using either `test_montecarlo_pd_controller.py` or `montecarlo_pd_controller_test.py`

simulation/sub8_montecarlo/nodes/example_verification.py
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env python
+'''Monte-Carlo controller verification
```
pemami4911: Move example usage of the VC class to controller_verify.py

jpanikulam: Removed the extraneous test for this, and changed example_verification into a plotting tool

simulation/sub8_montecarlo/montecarlo_tests/montecarlo_pd_controller.py
jpanikulam: This gets kind of weird, because this is a tool for integration-testing the controller, and it's hard to devise useful unittests for this without a controller that is assumed to work...

Hmm...

jpanikulam: I'd imagine having several of these, like

**test_pd_controller_convergence**
**test_pd_controller_max_q**
**test_pd_controller_orientation_constraints**

**test_RISE6dof_convergence**
**test_RISE6dof_max_q**
**test_RISE6dof_orientation_constraints**

Or something like that

jpanikulam:
Added two independent commits addressing the comments on the changes to Alarm Tests and to the core MonteCarlo tools.

jpanikulam:
Killing this PR to push to feature branch

PR #30, opened 2015-11-06
Introducing so3tools.

jnez71:
Adding version 1.1 of the so3tools library, as well as a unit test and a demo program to show an application of the library.

### DIFF
--- a/utils/sub8_ros_tools/so3tools/demo_so3tools.py
+++ b/utils/sub8_ros_tools/so3tools/demo_so3tools.py
@@ -1,0 +1,220 @@
+"""Demo of using so3tools in an attitude dynamics simulation and PD controller."""
+
+# --- DETAILS
+# The script below is divided into SETUP, SIMULATE, and DISPLAY. In the SETUP
+# section you can adjust simulation length, timestep, initial conditions, desired
+# state, mass properties, controller gains, and the display framerate. In the
+# SIMULATE section, you will find a verbose description of exactly what is done
+# to carry out the simulation. In short, it is an Euler forward integration, and
+# the controller applied is a simple PD controller. The simulation and controller
+# code is super short and neat thanks to so3tools. Once the simulation results
+# have been stored, the script then goes into DISPLAY where the terror that is
+# matplotlib is used to animate the results. You will notice that the animation is
+# slower than real time - this is because matplotlib is garbage. Remember that it is
+# only displaying results; the simulation has already been completed. In the animation,
+# rigid body orientation is represented as 3 perpendicular lines. Dotted lines show the
+# initial orientation, and dashed lines show the desired orientation. The animation loops
+# when it reaches the end of the data, which depends on what you set the duration to in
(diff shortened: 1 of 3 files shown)


utils/sub8_ros_tools/so3tools/demo_so3tools.py
```diff
@@ -0,0 +1,220 @@
...
+# You probably want to do this a little or else the lines stop appearing 3D to the eye.
+
+# --- AUTHOR
+# Jason Nezvadovitz
+
+################################################# IMPORTS
+
+# standard
+from __future__ import division
+import time
```
jpanikulam, line 29: Unused import F401

utils/sub8_ros_tools/so3tools/demo_so3tools.py
```diff
@@ -0,0 +1,220 @@
...
+# 3rd party
+import numpy as np, numpy.linalg as npl
```
jpanikulam, line 31: Multiple imports on one line. E401

utils/sub8_ros_tools/so3tools/demo_so3tools.py
```diff
@@ -0,0 +1,220 @@
...
+# Visually represent the body as three perpendicular lines, each defined by its endpoints:
+linlen = 0.5
+xline = np.array([[0,linlen],[0,0],[0,0]])
+yline = np.array([[0,0],[0,linlen],[0,0]])
+zline = np.array([[0,0],[0,0],[0,linlen]])
+body = np.concatenate((xline, yline, zline), axis=1)  # each column is a point in body
+
+# Initialize histories for recording the simulation:
+q_history = np.zeros((len(t_arr),4))
+roll_history, pitch_history, yaw_history = np.zeros(len(t_arr)), np.zeros(len(t_arr)), np.zeros(len(t_arr))  # for plotting something understandable
```
jpanikulam, line 83: Try to keep line length under 130

utils/sub8_ros_tools/so3tools/demo_so3tools.py
```diff
@@ -0,0 +1,220 @@
...
+    I_world = so3.apply(q, so3.apply(q, I).T)  # new I_world computed using new q
+    H = H + torque*dt  # new H from old H and torque
+    w = npl.inv(I_world).dot(H)  # new angular velocity computed using new I_world and new H
+
+################################################# DISPLAY
+
+fig = plt.figure()
+fig.suptitle('Orientation State Evolution', fontsize=24)
+
+if animate == True:
```
jpanikulam, line 157: Do 

```
if animate:
```

or 

```
if animate is True
```

instead of equality comparison (E712)

forrestv, line 157:
```
if animate is True:
```

is bad. I think pylint allows it so that there is some way of checking that a value is literally exactly `True`, but you'd pretty much never want to actually do that.

Use the first example.

PR #31, opened 2015-11-06
SO3TOOLS: Introducing so3tools, PEP8 Conformed

jnez71:


forrestv:
I said some of this last week and I suppose I'll say it again here:

This looks pretty overengineered. A lot of code. 80% of the operations that this supports will never be used - e.g., when are you ever going to "add" two "axles"? You should strive to create a sufficient, but minimal interface - one in which everything will be used and there is no redundancy. There should not be multiple ways to do the same thing. To quote The Zen of Python, "There should be one-- and preferably only one --[...] way to do it." Redundancy creates untested code paths which creates room for bugs.

The way that I recommended doing that was choosing a canonical representation (for which I believe quaternions to be the best choice) and then providing only operations on it and then conversion routines between it and every other type. By choosing one canonical representation and only allowing operations involving it, you decrease the amount of code paths by a factor of the number of representations you support.

Also, as clever as "axle" is, I'd recommend dropping it. Things are confusing enough as they are. There's no reason to add to the confusion with a term will be useless to Google/whatever. Choose your favorite from {Euler vector, rotation vector}.

Frankly, we got along fine with only `tf.transformations` and

```
def quat_to_rotvec(q):
    if q[3] < 0:
        q = -q
    q = transformations.unit_vector(q)
    angle = math.acos(q[3])*2
    axis = normalized(q[0:3])
    return axis * angle

def rotvec_to_quat(rotvec):
    return transformations.quaternion_about_axis(numpy.linalg.norm(rotvec), rotvec)
```

, and I don't see the point in adding nearly a thousand lines of new code to implement the same functionality. I'm sure that this was a great learning exercise and could function as a useful demo, but again, I don't really see the need.

Sorry if this seems a bit harsh. Tell me if I'm making any bad assumptions.

forrestv:
From your comments,

> The functions are generally representation independent; you can pass in a quaternion or rotation matrix or whatever valid SO3 quantity you want and high level operations are carried out correctly regardless.
> 
> SO3 quantity typing is not object-oriented. This means you can keep your quantities as simple arrays; you don't have to make all sorts of wacky class function calls to work on a state with SO3 quantities. For example, you could fit your favorite quaternion q into a state vector as easy as [blah,q] and all numpy array operations will still be usable.

I gleaned more of the motive behind this. However, I think that your intent is misplaced.

I already more or less addressed the first line, but I'll say that achieving that isn't very useful since you should always know what type of SO3 representation you're working with. You can't just completely forget, as each representation has some drawbacks and things you need to keep in mind.

For the second, you seem to have an irrational fear of OO. Your rationalization of being able to keep things as "simple arrays" to do "numpy array operations" with sounds nice until you realize that it rarely makes sense to add or elementwise multiply the components of a quaternion or rotation vector. So if you can't really use any array operations, why have your state in one big vector at all? I think you should take another look at the box-plus paper. The entire point of it is that you can get a local vector-space view of the neighborhood of a point on an arbitrary manifold. And yes, you can work with that as a vector, but not the original state. There's no disadvantage to using OO for your state - some kind of record structure to hold everything by name, a quaternion object to support operations on the orientation, etc. Lots of advantages there - the best is not having to deal with indices, which are be an easy place to introduce bugs. In fact, OO can help by forcing you to only interact with the state in a few safe ways.

jnez71:
Regarding your first comment:

I completely see where you're coming from. Under the hood, I strove for completion over minimalism. The generalized compatibility of every so3tools function with not just quaternions, but also rotation matrices and rotation vectors is functionality that we do not absolutely need. I had the mindset that it would just be nice if it was there, and of course nice to have the learning experience of developing it.

However, I disagree with the statement "80% of the operations that this supports will never be used." Or rather, I'd argue that it's an exaggeration. The demo, which encompasses basic dynamics simulation and a controller (both relevant things to the sub), makes highly justified use of 7 out of the module's 14 functions (counting function calls internal to so3tools). The number isn't the point though, it's the way things look on the user's end- and that is my biggest point to emphasize. For example, forward stepping the orientation state in the simulation is:

```
dq = so3.quaternion_from_axle(w * dt)
q = so3.plus(dq, q)
```

That code has a parallel style to:

```
dx = dv * dt
x = x + dx
```

I don't think I even need to explain how so. And really, you're right, it kind of boils down to a pedagogical tool / learning experience for those in the lab who play with these functions. But I personally feel like there is usefulness to the simplicity and clarity in making the code appear as shown above in actual sub code. As another example, the PD controller is:

```
q_err = so3.error(q, q_des)
w_err = w_des - w
torque = (kp * q_err) + (kd * w_err)
```

I can read those lines out loud and people would understand them. They might then say, "yeah but just grabbing an error between q and q_des is easier said than done." Well now it's easier done than said. (That was also an example of a case where I wanted to do element-wise list operations on what otherwise may have been a class object, since kp and kd are just lists).

Using only tf.transformations and orientation_helpers, the first line of the PD controller would have had to be:

```
q_err = quat_to_rotvec(transformations.quaternion_multiply(q_des, transformations.quaternion_inverse(q)))
```

There is nothing wrong with that. In fact, that is exactly what is done internal to so3tools. I emphasize to everyone that all so3tools does is package things in a more comprehensible way. It does not present anything new or clever that was not available in the combination of tf.transformations and orientation_helpers. In that regard, we by definition do not need this module. What I am arguing, however, is that "80%" of the module _does find use in making user code more comprehensible._

Regarding your second comment:

This is a misunderstanding (my fault). I very much support OO. My explanation of why I avoid OO in so3tools was poor though. What I meant to say was that so3tools is not meant to be our sub8 state class. I am all for us developing a neat and safe state class, and interacting with our state variables through it. However, this is not what so3tools is supposed to be or become. Rather, by leaving it as a bunch of general purpose functions, it could be used in anyone's code (even non-sub code) or in anyone's state class as simply compact and prettied up general purpose operations. To me this is not a strong argument though, and frankly I would have been okay with doing an OO so3tools instead. I just kinda picked a direction and stuck with it. I am open to suggestions for re-implementation.

I really appreciate the meaningful criticism so don't worry about sounding harsh.

Oh and regarding the name axle.. I will _reluctantly_ and I mean **reluctantly** give that up. Reluctantly. :P We'll go with rotation vector / rotvec to keep traditions and google searchability alive. I'll make those changes once we figure out what it is we intend to do with this module. Use it any way? Throw it out and just go commando with tf.transformations? Trim it down / remove the maximalist overhead? OO?

jnez71:
I have a couple more things to add in favor of so3tools, but then also a compromise that I think gives what both of us are looking for from this module.

First, coming back to your question "when are you ever going to add two axles [rotation vectors]?" Well, it is actually completely possible to keep orientation in the state as a rotation vector now. As an example, suppose for that PD controller, we stored our orientation as a rotation vector r rather than a quaternion q. All you would have to do is replace q with r in every location. As an example of code that actually becomes even simpler using r, the simulation code would be:

```
dr = w * dt
r = so3.plus(dr, r)
```

And we could still make use of so3.apply() when converting vectors and matrices to and from various frames. The code is always neat and similar in configuration because it's heavily abstracted. This is not bad nor is it necessarily good. It is a trade-off and something the rest of the team would need to speak up about in terms of what is wanted.

The other point is that centralizing our functions to a single source makes it easier to troubleshoot relevant problems, and once it's fixed, it gets fixed everywhere. This is obvious, and it's in effect why you made orientation_helpers. I would say though that so3tools takes that benefit even farther with the extra level of abstraction, since

```
q_err = quat_to_rotvec(transformations.quaternion_multiply(q_des, transformations.quaternion_inverse(q)))
```

won't be buried in the controller, planner, estimator, etc by each person who re-implements that line. Instead it will always just be so3.error().

We use tf.transformations like this too. There are probably a dozen functions in tf.transformations "we never use." So perhaps so3tools would seem better not as "sub code" but as a separate entity that the sub just imports?

That is one alternative, but the real alternative I want to suggest would be a new module called qtools. It will be so3tools with only quaternion compatibility. That means it will be just apply, plus, minus, error, rotvec_from_quat, and quat_from_rotvec, and each of those functions will also be smaller because they won't have to be able to handle all possible representations. This keeps what I like (the abstracting functions like plus, etc) and removes the over-engineered parts we don't need since we are of course keeping orientation in our state as a quaternion. What do you think of this?

utils/sub8_ros_tools/so3tools/so3tools.py
```diff
@@ -0,0 +1,557 @@
...
+    if shape[0] == 3:
+        # Check representation of x:
+        xrep = rep(x)
+        # Get rotation matrix if not rotmat:
+        if xrep == 'axle':
+            x = matrix_from_axle(x)
+        elif xrep == 'quaternion':
+            x = trns.quaternion_matrix(x)[:3, :3]
+        elif xrep == 'tfmat':
+            if t == 'tfmat':
```
jpanikulam, line 159: This is a very weird way of doing things

utils/sub8_ros_tools/so3tools/so3tools.py
```diff
@@ -0,0 +1,557 @@
...
+    """
+    angle = np.mod(npl.norm(r), 2 * np.pi)
+    if not np.isclose(angle, 0):
+        axlemat = crossmat(r / angle)
+        return np.eye(3) + (np.sin(angle) * axlemat) + ((1 - np.cos(angle)) * axlemat.dot(axlemat))
+    else:
+        return np.eye(3)
+
+
+def affine(x, t=None):
```
jpanikulam, line 466: This should use a more descriptive name

Suggestion: make_homogeneous

utils/sub8_ros_tools/so3tools/so3tools.py
```diff
@@ -0,0 +1,557 @@
...
+    """Returns the affine space form of a given array x. If x is
```
jpanikulam, line 467: Bullet points would make this easier to human-parse than paragraph form

utils/sub8_ros_tools/so3tools/so3tools.py
jnez71, line 159: How so? The idea is that regardless of the rep of x, I intend to apply it by doing a matrix multiplication (provably faster). So I only check for the other possible input types, since if they passed in a rotmat, no pre-conversion is necessary.

Or were you referring to the fact that I put all the code inside the if statement that checks the shape of P, rather than just having

```
if shape[0] != 3:
    raise ValueError('...')
-continue code from here-
```

Yeah, I'm not sure why I organized it that way. I can fix that.

utils/sub8_ros_tools/so3tools/so3tools.py
```diff
@@ -0,0 +1,557 @@
...
+    """Returns the skew-symmetric matrix form of a three element vector.
+    See: https://en.wikipedia.org/wiki/Cross_product#Conversion_to_matrix_multiplication
+
+    >>> a = [1,2,-3]
+    >>> b = [4,-5,6]
+    >>> print(np.cross(a,b) == crossmat(a).dot(b))
+    [ True  True  True]
+
+    """
+    return np.array( [[  0   , -v[2]  ,  v[1] ],
```
jpanikulam, line 514: Don't put spaces before/after parentheses (E201, E202)

utils/sub8_ros_tools/test_ros_tools/test_so3tools.py
```diff
@@ -0,0 +1,154 @@
...
+        M = trns.random_rotation_matrix()
+        self.assertEqual(so3.rep(M), 'tfmat')
+        self.assertEqual(so3.rep(M[:3, :3]), 'rotmat')
+        self.assertEqual(so3.rep(so3.get_axle(M)), 'axle')
+        self.assertEqual(so3.rep([5, 1, 0]), 'notSO3')
+        M[-1, -1] = 2
+        self.assertEqual(so3.rep(M), 'notSO3')
+
+    def test_apply(self):
+        P = [[1, 0, 1],
```
jpanikulam, line 23: Use some whitespace to separate out code-chunks

utils/sub8_ros_tools/so3tools/demo_so3tools.py
```diff
@@ -0,0 +1,220 @@
+"""Demo of using so3tools in an attitude dynamics simulation and PD controller."""
+
+# --- DETAILS
```
jpanikulam, line 3: Use multiline strings for multiline comments

jpanikulam:
Some general notes:
- There is a lot of comparison overhead - I cannot think of a case where someone using this does not already know what representation they are using. This could be solved with OO, (You could even overload "+" and "-") or by implementing seperate functions for each type. I am partial to the OO approach.
  -  The useful applications of this are in simulation, motion planning, state estimation, and the controller. Most of these are very high frequency, milliseconds count

In the future:
- Be sure to prepend your commits with a topic
- Try to submit pull requests in chunks (900+ lines of very mathematical code is a lot to review)
- Doctests are very hard to read, and don't provide great insight into how some set of code is meant to be used. [Here's a good article](http://bemusement.org/doctests-arent-code) about it that provides more insight than "unittest is magically better".
- Long inline comments are hard to read
- There is no need to abbreviate function or variable names, typing 6 more letters costs negligible time
- In the future, test your code on an actual Ubuntu machine running the real Sub8 stack. It is not sufficient to assume that Python's handy cross-platform capabilities will carry us through
- If you need to write more than a few lines to explain what is going on, consider making a wiki entry that allows you to use real markdown formatting.

jpanikulam:
I am getting these errors:

```
======================================================================
ERROR: test_plus (test_so3tools.TestSo3Tools)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/jacob/repos/Sub8/utils/sub8_ros_tools/so3tools/test_so3tools.py", line 54, in test_plus
    q1 = trns.quaternion_from_matrix(R1)
  File "/opt/ros/indigo/lib/python2.7/dist-packages/tf/transformations.py", line 1208, in quaternion_from_matrix
    if t > M[3, 3]:
IndexError: index 3 is out of bounds for axis 0 with size 3

======================================================================
FAIL: test_angle_axis (test_so3tools.TestSo3Tools)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/jacob/repos/Sub8/utils/sub8_ros_tools/so3tools/test_so3tools.py", line 114, in test_angle_axis
    self.assertTrue(np.allclose(myAngle, myAngle2))
AssertionError: False is not true

======================================================================
FAIL: test_error (test_so3tools.TestSo3Tools)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/jacob/repos/Sub8/utils/sub8_ros_tools/so3tools/test_so3tools.py", line 85, in test_error
    self.assertTrue(np.allclose(err_near, axle_target_near - axle_current, atol=1e-02))
AssertionError: False is not true

======================================================================
FAIL: test_get_axle (test_so3tools.TestSo3Tools)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/jacob/repos/Sub8/utils/sub8_ros_tools/so3tools/test_so3tools.py", line 105, in test_get_axle
    self.assertTrue(trns.is_same_transform(M, trns.quaternion_matrix(q)))
AssertionError: False is not true

----------------------------------------------------------------------
Ran 14 tests in 0.049s

FAILED (errors=1, failures=3)
```

jnez71:
Alright, here is a summary of the changes that I am going to make:
- Don't contain the bulk of code inside a quick-check if statement.
- Change affine() to make_affine() or make_homogeneous()
- Use bullets in docstrings when listing things
- Remove the accidental extra spaces found in crossmat()
- Add white space to the unit test (when I made the test, I just copied my docstring stuff and cared little about formatting)
- Use multiline strings for multiline comments, or rather just make a wiki for this type of lengthy documentation
- Write commit messages according to our wiki rules
- **Turn so3tools into qtools (or a better name): remove the representation-checking overhead used everywhere and leave it as basically a combination of geometry_helpers and plus, minus, error, and apply.** (I am under the assumption that you guys see some value in those functions).
- The demo script will be placed somewhere else and a link will be provided in the qtools documentation, since the demo is written more like a tutorial than code.

Jake: the errors you are getting may be related to the fact that I just import transformations while for you it comes from tf. I do not get those errors running the unit test on my computer. It looks like each error you got was when a call to transformations was made.

utils/sub8_ros_tools/so3tools/so3tools.py
```diff
@@ -0,0 +1,557 @@
...
+    >>> M2 = trns.rotation_matrix(np.pi, [0,1,0])
+    >>> print(trns.is_same_transform(M1, M2))
+    True
+
+    """
+    angle = np.mod(npl.norm(r), 2 * np.pi)
+    if not np.isclose(angle, 0):
+        return trns.quaternion_about_axis(angle, r / angle)
+    else:
+        return np.array([1, 0, 0, 0])  # unit real number is identity quaternion
```
forrestv, line 443: Correct me if I'm wrong, but I think that the convention that transformations.py uses is XYZW, so this isn't the identity quaternion. At least on my system:

```
>>> from tf import transformations
>>> transformations.quaternion_about_axis(0, [1, 0, 0])
array([ 0.,  0.,  0.,  1.])
```

jnez71:
Yes, there is a version discrepancy in transformations.py.
I am using version 2015.07.18 found here: http://www.lfd.uci.edu/~gohlke/code/transformations.py.html
The version used in ROS tf is 2009.04.18 found here: https://github.com/ros/geometry/blob/indigo-devel/tf/src/tf/transformations.py

I don't intend to make so3tools (or the almost complete simpler version, oritools) into Sub8 official software anymore. If desired, we can use oritools as an external dependency. You can find the latest code here: https://github.com/jnez71/Orientation_Library

PR #27, opened 2015-11-02
Add goal client and start of mission system

zachgoins:
I am pull requesting now because I wanted to split it up. This is the initial implementation of a goal server for use in planning missions and issuing positions of the sub. At the moment only the client side of the system is done, but I have included a simple server so you can run all tests and verify output. I am going to add info to the wiki as I confirm implementation.

PR #32, opened 2015-11-07
Add goal client and helper geometry tools 

zachgoins:
Supersedes #27


jpanikulam:
Last comment I promise:

In the function "relative_move" in goal_client.py, there are a bunch of arguments that aren't used in the actual function - are they doing something weird? It seems like the waypoint creation uses data that was created elsewhere

zachgoins:
My bad. That was leftover after refactoring. It is fixed now.

jpanikulam:
Approved: Jacob Panikulam

PR #33, opened 2015-11-08
Monte-Carlo: Add simple controller verification

jpanikulam:
Committing to a feature branch now, until sufficiently complete (so as to add to master)


jpanikulam:
Addressing both of those comments and the alarm severity requirement. The cause of the change was that 0.2 seconds was the smallest time that I found always worked.

jpanikulam:
Addressed

simulation/sub8_montecarlo/sub8_montecarlo_tools/controller_verify.py
```diff
@@ -0,0 +1,200 @@
...
+        self.target_state = pose[0]
+
+    def odom_cb(self, msg):
+        time_of = rospy.Time.now()
+        if (time_of - self.last_sample) < self._sampling_period:
+            return
+        else:
+            self.last_sample = time_of
+
+        pose, twist, _, _ = sub8_utils.odometry_to_numpy(msg)
```
zachgoins, line 101: I like this method of grabbing messages. Fancy Shmancy

PR #40, opened 2015-11-30
SIM: Implement DVL, IMU sensors

Annhiluc:
Implemented the DVL and IMU sensors to publish information in the simulation - DVL sensor information is in a new ROS message (DVLSim) and IMU sensor information is in ROS imu message.

@jpanikulam


simulation/sub8_simulation/sub8_sim_tools/physics/vehicle.py
```diff
@@ -121,6 +171,62 @@ def publish_pose(self):
...
+            frame_id='/world'
+        )
+
+        linear=geometry.Vector3(*linear_vel)
+        angular=geometry.Vector3(*angular_vel)
+
+        noise = np.random.random(3)
+
+
+        covariance = [noise[0], 0., 0.,
```
forrestv: You appear not to know what covariance is... It's a description of the noise present, not the actual noise signal.

You should be adding `noise` to one of the actual measured quantities (`angular` or `linear`), not using it as the covariance. The covariance should be a constant diagonal matrix.

Also, an accelerometer measures acceleration, not linear velocity.

I wouldn't bother filling in the orientation. An IMU doesn't directly measure orientation, so it's kind of weird that it's in the IMU message at all. The old software didn't fill it in or use it anywhere. [The message](http://docs.ros.org/api/sensor_msgs/html/msg/Imu.html) says

> If you have no estimate for one of the data elements (e.g. your IMU doesn't produce an orientation estimate), please set element 0 of the associated covariance matrix to -1

simulation/sub8_simulation/sub8_sim_tools/physics/vehicle.py
```diff
@@ -149,4 +255,42 @@ def step(self, dt):
...
+
+
+    def get_dvl_vel(self):
+        """Returns array of 4 components of the vehicle's velocities based off of the dvl ray vectors"""
+        dvl_ray_1 = np.array([0.866, -.5, -1])
+        dvl_ray_2 = np.array([0.866, .5, -1])
+        dvl_ray_3 = np.array([-0.866, .5, -1])
+        dvl_ray_4 = np.array([-0.866, -.5, -1])
+
+        vel_sub_world = self.body.vectorFromWorld(self.body.getRelPointVel(self.dvl_position))
```
forrestv: This is definitely a body-frame quantity, so it's probably better named `vel_sub_body`. It's a bit more subtle than that, though, since it's not the velocity of the sub, but the velocity of the DVL ... so `vel_dvl_body` is probably better yet.

forrestv:
Looks good, except for previous comments!

simulation/sub8_simulation/sub8_sim_tools/physics/vehicle.py
```diff
@@ -53,11 +71,43 @@ def __init__(self, world, space, position):
...
+            self.body.addRelTorque(np.array([-500,0,0]))
+        elif(data.data == 'h'):
+            self.body.addRelTorque(np.array([500,0,0]))
+        elif(data.data == 't'):
+            self.body.addRelTorque(np.array([0,500,0]))
+        elif(data.data == 'g'):
+            self.body.addRelTorque(np.array([0,-500,0]))
+        elif(data.data == 'r'):
+            self.body.addRelTorque(np.array([0,0,500]))
+        elif(data.data == 'v'):
```
zachgoins, line 107: Will there ever be a time we will need to move in more than one axis, rotate and translate, rotate in multiple axes, etc at once? using elif would not allow for it.

jpanikulam:
Hey Annie, could you run

```
flake8 --max-line-length=130 --ignore=E201 ./
```

On your changes, and make the changes such that it is PEP8 compliant?

simulation/sub8_simulation/sub8_sim_tools/physics/vehicle.py
```diff
@@ -5,10 +5,13 @@
 from sub8_ros_tools import rosmsg_to_numpy, normalize
 from sub8_sim_tools.physics.physics import Box
 from sub8_simulation.srv import SimSetPose, SimSetPoseResponse
+from sub8_simulation.msg import DVLSim
```
jpanikulam: Is the file for DVLSim commited?

Also - put it in sub8_msgs and make it [like this](https://github.com/uf-mil/software-common/blob/master/uf_common/msg/VelocityMeasurements.msg), so we can use the old UKF with minimal adjustment

simulation/sub8_simulation/sub8_sim_tools/rendering/visualization.py
```diff
@@ -48,34 +53,63 @@ def on_draw(self, event):
         self.world.draw(self.view)
 
     def on_key_press(self, event):
+        # Process events on a specific object
+        if(event.text.lower() == 'i'):
```
jpanikulam, line 57: Why make these ifs instead of elif? Only one key can be pressed at a time and intercepted

PR #41, opened 2015-12-01
TGEN: New rostests, rosparams, B matrix service for thruster mapper

pemami4911:
Bunch of new additions to the TGEN, in short: 
- B matrix service exposed by the thruster mapper and called by the TGEN
- rostests for inter-node testing with the thruster mapper
- Moved TGEN configuration parameters to rosparams set in a launch file
- Other minor misc fixes/changes


command/sub8_alarm/package.xml
```diff
@@ -1,14 +1,17 @@
 <?xml version="1.0"?>
-<package>
+<package format="2">
```
jpanikulam, line 2: What does this do?

gnc/sub8_trajectory_generator/package.xml
```diff
@@ -13,14 +13,11 @@
...
-  <build_depend>roscpp</build_depend>
-  <build_depend>OMPL</build_depend>
-  <build_depend>sub8_msgs</build_depend>
-  <build_depend>cmake_modules</build_depend>
-  
-  <run_depend>geometry_msgs</run_depend>
-  <run_depend>roscpp</run_depend>
-  <run_depend>OMPL</run_depend>
-  <run_depend>sub8_msgs</run_depend>
+  <depend>geometry_msgs</depend>
```
jpanikulam, line 16: Oh I see what it does. Slick!

simulation/sub8_simulation/sub8_sim_tools/rendering/world.py
```diff
@@ -256,6 +256,7 @@ def __init__(self):
         - Add ros-camera (How to handle rigid-body offsets?)
         '''
         self.entities = []
+        self.my_lights = []
```
jpanikulam, line 259: Sweet baby jesus we need to figure out a better merging strategy

utils/sub8_msgs/srv/BMatrix.srv
```diff
@@ -0,0 +1,2 @@
+---
+float64[] B
```
jpanikulam, line 2: The thruster mapper internally changes B when a thruster fails - perhaps we should change this in light of both the TGEN and the mapper using the same matrix.

Ignore for now and let's talk about it on Weds

gnc/sub8_thruster_mapper/nodes/mapper.py
```diff
@@ -82,6 +85,10 @@ def get_thruster_wrench(self, position, direction):
         wrench_column = np.hstack([forces, torques])
         return np.transpose(wrench_column)
 
+    def get_b_matrix(self, srv):
```
jpanikulam, line 88: We should probably thread_lock this lest we accidentally send B while it is being generated. 

That, or only start advertising the service once we've generated B for the first time

pemami4911, line 88: Hm yeah, it probably should be thread locked, and the "thruster_failure" alarm handler should alert the TGEN that it needs to request a new B as soon as the thruster mapper is done re-computing it.

jpanikulam:
Final question: Does this fix the build bug where you'd sometimes have to run catkin_make twice due to some header not building first?

Because I am [encountering it in CI](https://semaphoreci.com/jpanikulam/sub8/branches/control/builds/8) and it is breaking my heart

jpanikulam:
We will address thread-locking and alarm stuff in a future PR

PR #44, opened 2015-12-03
MISC: Fixes for TGEN Catkin build error

pemami4911:
Build on Semaphore repeatedly passing with both `catkin_make -j1` and `catkin_make -j8`


jpanikulam:
Victory at last!

PR #47, opened 2015-12-18
TGEN: Improved distance function, refactoring, and new tests

pemami4911:
In preparation for end-to-end testing of the TGEN,
I updated the distance function to a more sophisticated
metric that uses a weighted euclidean distance between
two states. See the implementation for more details.

Also did some minor refactoring to the Sub8StateSpace.

Updated unit and integration tests to reflect all changes.


gnc/sub8_trajectory_generator/launch/trajectory_generator_params.launch
```diff
@@ -16,15 +16,20 @@
 	<param name="propagation_step_size" value="2.0" type="double"/>
 	<param name="min_control_duration" value="1.0" type="double"/>
 	<param name="max_control_duration" value="100.0" type="double"/>
+
```
jpanikulam, line 19: Minor nitpick: Explicitly mention units -- Despite it being assumed that we're an all metric shop, we should leave nothing to chance.

[Reason](https://en.wikipedia.org/wiki/Mars_Climate_Orbiter)

gnc/sub8_trajectory_generator/src/sub8_state_space.cc
```diff
@@ -18,30 +21,58 @@ void Sub8StateSpace::freeState(State* state) const {
...
+  // compare position, velocity, and angular velocity
+  for (unsigned int i = 0; i < 9; ++i) {
+    if (fabs(state_a(i) - state_b(i)) > error) {
+      return false;
+    }
+  }
+
+  // compare orientation
+  // dot of two similar orientations should be 1
+  if (fabs(state_a.segment(9, 4).dot(state_b.segment(9, 4))) <
```
jpanikulam, line 60: This might be risky business: Since -q and q represent the same rotation for any quaternion, q, and q.dot(-q) = -1. 

This would yield false negatives in the case that q_a = -q_b.

I think your dot product idea is a good one, though. Alternative is compute acos((q_a \* q_b.inv()).w), but it is probably not worth the trig and quaternion inverses when all you want is "are these the same" and not "how different are these"

gnc/sub8_trajectory_generator/src/sub8_state_space.cc
```diff
@@ -18,30 +21,58 @@ void Sub8StateSpace::freeState(State* state) const {
...
 
-bool Sub8StateSpace::equalStates(const State* state1,
-                                 const State* state2) const {
-  // TODO
-  // Do we want to define some error-vector
-  // s.t. State1 == State2 if State1 is within += error-vector of State2?
   return true;
+}
+
+void Sub8StateSpace::enforceBounds(State* state) const {
```
jpanikulam, line 68: Does this always get called, or only when the bounds are violated?

If always, never mind. If only on violation, keep in mind that norm([0.99, 0.99, 0.99, 0.99]) = 1.98, but would not violate the bounds

gnc/sub8_trajectory_generator/src/sub8_state_space.cc
jnez71, line 60: I think his use of fabs() prevents that problem.

jpanikulam, line 60: Good catch, nix my comment

gnc/sub8_trajectory_generator/src/sub8_state_space.cc
pemami4911, line 68: This method is always called. It only does anything if the conditional at line 73 is true, however.

gnc/sub8_trajectory_generator/launch/trajectory_generator_params.launch
pemami4911, line 19: Will address this in a future commit, good call

jpanikulam:
The world is at ease!

PR #48, opened 2015-12-19
Montecarlo: Add montecarlo tools

jpanikulam:
This is a set of tools for testing things with random initial conditions, with special attention paid to controllers. With this, the montecarlo branch will be closed.


gnc/sub8_controller/include/sub8_controller/pd_controller.h
```diff
@@ -47,8 +48,8 @@ class PDController {
   // Control Parameters
   unsigned int trans_history_length = 75;
   unsigned int angle_history_length = 75;
-  controller_deque translation_error_history;
-  controller_deque orientation_error_history;
+  ControllerDeque translation_error_history;
+  ControllerDeque orientation_error_history;
 
   // TODO: Make these ROSParams
   double kp_trans = 30;
```
jnez71, line 55: In the future, there should be a tunable gain for each degree of freedom, like [kp_x, kp_y, kp_z], [ki_x, ki_y ...
These gains will be in body frame so that we can tune for performance differences like the difficulty in strafing compared to moving forward.

If R represents the world_from_body rotation, then something along the lines of

```
k_world = R * diag(k) * R.T
```

will be needed to put them in world frame so they can be used like

```
controller_output = k_world * error
```

pemami4911:
Pulled changes locally and verified that it builds and executes tests. Unless you want to make any changes to address Jason's comments, it looks good to go.

jpanikulam:
Adding Jason's comments to the TODO's for future controllers. 

The demo PD in this PR is not to be run on the real sub, and is almost entirely for unit-testing the Montecarlo verifier (Note one of the unit-tests requires the controller to fail). As such, I don't want to burn too much time improving it. 

Would welcome a pull-request improving this guy.

jnez71:
I know, the comments made that clear. To-do notes are sufficient.

PR #49, opened 2015-12-23
Simulation: Add widgets, fix numerous issues

jpanikulam:
Bugfixes:
    - Fix monte-carlo issue where sometimes running faster than real-time would cause garbage behavior (You can now run as fast as your computer permits)
    - Fixed some crashes on keyboard press
    - Fixed some crashes related to sub's last force being known
    - Fixed crashes due to typo in visualization
    - Bugs encountered by Patrick
    - Move everything to make_header
    - Fix now-broken PD controller (as a result of sim changes)

Improvements:
    - Add gravitational bias to IMU
    - Add simulator "Widget" for the Sub
    - Add simulator option "demo_mode" particularly for @pemami4911 , demo mode spools up the demo pd controller purely so that it can be used to test the trajectory generator while the sophisticated controllers are still in the oven
    - The demo pd controller does not support waypoints, a PR later this morning will add this feature for demo mode
    - Clean up vehicle sensor publishers
    - Odometry now publishes in the correct frames

Optimizations:
    - tbh I forgot what I did

### DIFF
--- a/gnc/sub8_controller/cfg/pd_controller.cfg
+++ b/gnc/sub8_controller/cfg/pd_controller.cfg
@@ -1,16 +1,16 @@
 #! /usr/bin/env python
-from dynamic_reconfigure.parameter_generator_catkin import *
+from dynamic_reconfigure.parameter_generator_catkin import ParameterGenerator, double_t, int_t
 
 gen = ParameterGenerator()
 
-gen.add("kp_trans", double_t, 0, "kp_trans", 30., 0.0, 30)
-gen.add("kd_trans", double_t, 0, "kd_trans", 13.5, 0.0, 30)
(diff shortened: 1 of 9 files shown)


jpanikulam:
@tessbianchi

tessbianchi:
Looks good to me!

jpanikulam:
Because I realized it looks odd: Jason and I discussed his comments offline. Buoyancy issue fixed, other things are known issues that will be addresses in future iterations. 

And by offline I mean on the internet, but in a different place on the internet.

jnez71:
This is fine, but it doesn't seem like the buoyancy issue did actually get fixed in this pull.
https://github.com/uf-mil/Sub8/blob/master/simulation/sub8_simulation/sub8_sim_tools/physics/physics.py#L207
Perhaps you missed including that diff?

PR #50, opened 2016-01-04
ALARMS: Add default 'all-is-well state'

rleyva:
Small change for the alarm system. States will now go from 0-3. 0 being the highest priority alarm, and 3 will the lowest priority state that denotes that no issues have occurred.

### DIFF
--- a/command/sub8_alarm/sub8_alarm/alarm_helpers.py
+++ b/command/sub8_alarm/sub8_alarm/alarm_helpers.py
@@ -20,7 +20,7 @@ class AlarmBroadcaster(object):
         # In the future, there will be some background bookkeeping
         self.alarms = []
 
-    def add_alarm(self, name, action_required=False, severity=2, problem_description=None, parameters=None):
+    def add_alarm(self, name, action_required=False, severity=3, problem_description=None, parameters=None):
         '''Factory method for creating a new alarm'''
         new_alarm = AlarmRaiser(
             alarm_name=name,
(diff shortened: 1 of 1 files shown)


rleyva:
@jpanikulam

jpanikulam:
Looks good.

@pemami4911 @rleyva  -- I'm imagining alarms2.0 having some "Is the sub ready to do XYZ", and have XYZ map to a list of services that must be in an operational state. Do you guys see merit to that? Ideas?

pemami4911:
@rleyva Can you update the relevant sections of the System Alarms wiki with these changes? 

When would a sub-system raise an "all-is-well" alarm? 

@jpanikulam Mmm maybe, but my initial thought is that the alarm system should only care about when some of the sub's sub-systems go down or malfunction; having it do something like keep track of which services are up and which aren't up yet is going to make it bloated. Do you want a separate high-level state machine that will only put the sub in an operational mode if all of the necessary services are up? (Just as an aside, the rqt GUI is useful for monitoring which services are up and which are down, and for sending dummy data to them)

jpanikulam:
@pemami4911 @rleyva

Good point. I do think we need some capacity to determine when the sub is in a "ready" state, and to my knowledge pure rospy stuff won't give us what we need. 

In alarms2.0, we should build out a tool - some kind of System Reports package - that subsumes both alarms and "I'm ready" messages. 
- Jake

rleyva:
@pemami4911 I took care of updating the wiki the other day. 

And @jpanikulam, rospy won't take care of this natively. I think this can be done very easily by simply giving the alarm system a list of nodes that need to be up and running before entering any other state. We could probably even segment the necessary nodes into separate groups and entering states based on that.

jpanikulam:
Let's conclude with a plan. Ralph, do you want to add such a feature to
your docket? I think it is low-priority behind perception stuff, but a
feature that an ideal sub would have.

rleyva:
I am finalizing perception hook-ins for the alarm system at the moment, but once I am done with that I will take care of this.

PR #51, opened 2016-01-05
SLAM: Add simple monocular tracking

jpanikulam:
Implement C++ utilities for monocular tracking, and a demo node that tracks frames in a video.

This is a dev version, and is not expected to pass a Semaphore build (I have not yet added dependencies to the build script). It has become large enough that an early pull-request is necessary to head off a massive diff in the future.

See slam.h for an "index" of useful functionality.


jpanikulam:
@tessbianchi

gnc/sub8_slam/src/slam.cc
```diff
@@ -0,0 +1,254 @@
...
+
+      std::cout << "PNP Error: " << pnp_error_amt << std::endl;
+      slam::draw_reprojection(render_frame, map_visible, pose_pnp, intrinsics);
+
+      // Magic number -- TODO why is this so large compared to 100.0?
+      if (pnp_error_amt < 2500.0) {
+        slam::Point3 pt(pose_pnp.translation);
+        camera_locations.push_back(pt);
+
+        slam::Frame key_frame(pose_pnp, new_feature_ids, new_feature_locations);
```
jpanikulam, line 224: The idea here (in the next iteration) is to initialize a new keyframe (and associated map section) when the reprojection error becomes too high. That is, we'd backtrack `k` frames from the time when pnp_error exceeded some threshold.

The blocking issue for this is that reprojections from PnP pose solutions are flawed somehow. I don't think switching to sparse model-based alignment is going to fix it yet.

jpanikulam:
Should we ravel everything into a single class, since the constant matrix, intrinsics, is getting passed around a lot?

Should frame objects carry the whole image with them? (This could grow in memory use super fast, maybe just keyframes, maybe not at all?)

tessbianchi:
I don’t think that's a good reason to bring a bunch of classes together, couldn’t you just define them somewhere, like in slam.h or no? 
If not I still think you should pass them around because I think the code will get out of control if you don't separate the parts early on.

I don't think you should have an array of images and an array of frames holding images, if this is what you are doing. Maybe a reference to the image?

jpanikulam:
Hmm. Will have to read camera calibrations from ROS on startup. We'll revisit the frame thing, I'm not yet convinced that we need to carry the full image for historical keyframes

Also check it out, Semaphore is passing

tessbianchi:
Yay! Nice!

tessbianchi:
Approved

PR #52, opened 2016-01-11
BUILD: Add install.sh, replacing the Semaphore one

jpanikulam:
- Replaces the script that lives on Semaphore
  - From now on, Semaphore will run this script

Reason:
    - We need to be able to track changes in dependencies
    - We need to be able to run/update our own machines w/o looking at
      Semaphore
    - It should be easy to see dependency changes _in pull-requests_

Note:
     - All historical builds will now fail
     - Any new PR will have to merge master to build properly in Semaphore

I think the route of straight-up replacing the old build script is much cleaner than leaving an option in Semaphore for a little while.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1,4 +1,18 @@
 #!/bin/bash
+if [ $# -ge 1 ]; then
+    DEPS_DIR="$1"
+    CATKIN_DIR="$1"/catkin_ws
+elif [ $# -ge 2 ]; then
+    DEPS_DIR="$1"
+    CATKIN_DIR="$2"
+else
+    DEPS_DIR=~/repos/sub_dependencies
+    CATKIN_DIR=~/repos/catkin_ws
+fi
+
+echo "Installing dependencies in $DEPS_DIR"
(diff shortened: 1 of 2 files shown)


install.sh
```diff
@@ -0,0 +1,158 @@
...
+    set -e
+    # Let's force ourselves to stay at the latest version
+    # TODO: python -c "import vispy; print vispy.__version__" -> Compare to what's available
+    # TODO: Uninstall old Vispy if it is out of date
+    git clone -q https://github.com/vispy/vispy.git
+    cd vispy
+    sudo python setup.py develop
+else
+    # How's that for a hack!? (Figure out location of Vispy)
+    cd "$(dirname `python -c  "import vispy; print vispy.__file__"`)/.."
```
pemami4911, line 71: nice lol

jpanikulam:
@pemami4911 Added, along with documentation for setting directories. It turns out that doing -c/-d is annoyingly complicated in bash, so for now we're just doing positional arguments.

pemami4911:
Looks good!

PR #53, opened 2016-01-20
SIM: Add shader manager to change entity shaders

tessbianchi:
You can now switch out the shaders of certain entities which got rid of all the annoying debug statements in the simulation. I included in the file about how to extend off my code, and write your own shader manager.

### DIFF
--- a/simulation/sub8_simulation/nodes/simulate.py
+++ b/simulation/sub8_simulation/nodes/simulate.py
@@ -10,6 +10,7 @@ import rospy
 
 
 class SimWorld(rendering.Canvas):
+
     def __init__(self):
         rospy.init_node('simulator')
         self.clock_pub = rospy.Publisher('/clock', Clock, queue_size=1)
@@ -25,11 +26,13 @@ class SimWorld(rendering.Canvas):
         # Physics world starts with such a plane
         self.rendering_world.add_plane((0.0, 0.0, 1.0), 100.0, 100.0, color=(0, 0, 255, 170))
 
-        self.rendering_world.add_mesh(Transdec, (0.0, 0.0, 0.0),
-                                      orientation=None, color=(155, 155, 100), shininess=3.0)
+        self.rendering_world.add_mesh(Transdec, (0.0, 0.0, 0.0), orientation=None, color=(155, 155, 100), shininess=3.0)
         self.physics_world.add_entity(Mesh, (0.0, 0.0, 0.0), 10., Transdec)
 
-        super(self.__class__, self).__init__(self.time_acceleration, show_window=self.draw, physics_dt=self.physics_dt)
+        self.rendering_world.add_point_light((0.0, 1.0, 0.0), (0.0, 0.0, 0.0))
+
+        super(self.__class__, self).__init__(
+            self.time_acceleration, show_window=self.draw, physics_dt=self.physics_dt)
 
         self.view = np.array([
             [1.,     0.,      0., 0.],  # noqa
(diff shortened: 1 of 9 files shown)


simulation/sub8_simulation/nodes/simulate.py
```diff
@@ -23,13 +24,16 @@ def __init__(self):
         self.entity_pairs = []
 
         # Physics world starts with such a plane
-        self.rendering_world.add_plane((0.0, 0.0, 1.0), 100.0, 100.0, color=(0, 0, 255, 170))
+        self.rendering_world.add_plane(
```
jpanikulam: Extra formatting?

simulation/sub8_simulation/sub8_sim_tools/rendering/world.py
```diff
@@ -17,18 +17,21 @@ def set_debug(self, debug=True):
...
-    def __init__(self, mesh, position=(0.0, 0.0, 0.0), orientation=None, color=(255., 0., 0., 255.0), faces=None):
+    def __init__(self, mesh, position=(0.0, 0.0, 0.0), orientation=None, color=(255., 0., 0., 255.0), faces=None,
+                 shader_manager=None):
         '''Orientation must be a 4x4 rotation matrix'''
         self.position = np.array(position, dtype=np.float32)
         if orientation is None:
             self.orientation = np.eye(4)
         else:
-            assert orientation.shape == (4, 4), "Orientation must be a 4x4 numpy array"
+            assert orientation.shape == (
```
jpanikulam: Extra formatting?

simulation/sub8_simulation/sub8_sim_tools/rendering/world.py
```diff
@@ -100,18 +103,20 @@ def make_buffer(self, mesh):
...
         sphere_mesh = geometry.create_sphere(
             rows=int(radius * 30),
             cols=int(radius * 30),
             radius=radius,
             **kwargs
         )
+
         sphere_buffer, faces = self.make_buffer(sphere_mesh)
         super(self.__class__, self).__init__(sphere_buffer, faces=faces, position=position, color=color)
+        shader_manager.register_lighting_shader(self)
```
jpanikulam, line 118: Would it be smoother to do registration inside of the rendering_world, when the object is created, instead of registering in each entity's constructor?

Ex:

```
def add_sphere(*args):
    sphere = Sphere(*args)
    self.shader_manager.register_lighting_shader(sphere)
```

Instead of each entity having to keep track of the shader manager?

tessbianchi, line 118: Yeah I thought that, but I think it could be useful later for the entity to have reference to the shader manager, so that if during runtime we want to change the shader then we can change the draw method of that entity to have it derregsiter from one shader_manger and register with another manager under certain conditions.

simulation/sub8_simulation/sub8_sim_tools/shaders/shader_manager.py
```diff
@@ -0,0 +1,70 @@
...
+
+    def __init__(self):
+        self.entities = []
+        self.light_manager = LightManager()
+
+    def register(self, entity):
+        if entity in self.entities:
+            raise ValueError("ERROR: You've registered one entity with more that one manager, this is not allowed.")
+        else:
+            self.entities.append(entity)
```
jpanikulam, line 59: Don't worry about his now, but if we have trouble deleting objects, we should make this a **weakref**

simulation/sub8_simulation/sub8_sim_tools/rendering/world.py
jpanikulam, line 118: Oh I see. My thinking is tha tit would be unusual for an Entity to change its own shader, but pretty normal for the world to change every Entity's shader at once when switching from camera to sonar rendering.

Not a breaking issue, just a thought

tessbianchi:
fin.

PR #57, opened 2016-02-18
TGEN: Alarm fix, README updates

pemami4911:
Some minor things: 
- Quick alarm bug fix for the TGEN
- Added extra state validity check for setting the problem definition, in case the start or goal state is placed on top of an obstacle & updated tests for the tgen_manager
- Updated the README so it now includes sample code for how to call the TGEN (In C++). Adapting this code to Python should be straightforward.


jpanikulam:
Accidental close, ignore.

PR #58, opened 2016-02-27
CAMERAS: Add drivers for all onboard cameras

rleyva:
@jpanikulam


jpanikulam:
Looks good overall, a few things
- There are extraneous .DS_STORE files
- It looks like camera1394 is an external ros package, you mentioned before you had some edits you wanted to make, but we should explicitly document them for posterity
- There are some launch files in camear1394 that look like they are Sub8 specific. We should shove them into sub8_launch (I wouldn't be opposed to a sub8_launch/launch/vision folder).
- I couldn't get the stereo.launch folder to run on Friday on the sub, does it work? If so, add a default for the manager argument
- Can you add a launch file for sending up stereo point cloud? I know we intend to port everything to CUDA, but we still need a usable interim solution
- It's not building (Check the Semaphore), add dependencies to install.sh

rleyva:
I’ll fix all of these things later tonight. I’m aware of the build issues, if I am not mistaken I have to change the permissions on the .cfg files to executables and everything should be okay. The single change is already documented in the wiki, but going forward there will be some other changes the I will make for the dc1394_reset_bus command so removing the package might not be an option unless I can get m fixes on the main repo of libdc1394.

jpanikulam:
Hey Ralph - any movement on this?

rleyva:
Taking care of it now... Will report back soon.

rleyva:
@jpanikulam So I fixed the camera issues, along with with a host of other issues that we had with the install script. The reason for all of the commits is that the ROS version on Semaphore was missing a bunch of stuff that comes with desktop-full. Look through everything and make sure that it's a-okay. The most reason successful build on Semaphore should contain all of the fixes for the cameras and install script. (I'll make sure to check the dependencies that we need ahead of time next time)

jpanikulam:
Looks great, thanks for the quick attention. 

Don't worry about needing to do many commits to make Semaphore work. I deal with it by having a seperate semaphore on my fork. 

In the future, if you want to make a commit that fixes the previous commit, you can do

git commit --amend
git push -f origin <branch>

(Note you're only force pushing to your fork, which is ok).

drivers/camera1394/launch/downward.launch
```diff
@@ -0,0 +1,50 @@
...
+    <node pkg="nodelet" type="nodelet" name="camera_nodelet_manager"
+          args="manager" />
+
+    <!-- Left/Right Camera Driver Nodelets -->
+
+    <!-- DOWNWARD CAMERA-->
+    <node pkg="nodelet" type="nodelet" name="down_camera1394_nodelet"
+          args="load camera1394/driver camera_nodelet_manager">
+          <param name="guid" value="00b09d01009ce76f"/>
+	  <param name="camera_info_url" value=""$(find camera1394)/calibration/${NAME}"/>
```
jpanikulam, line 12: Hey Ralph, this line seems to break the launch file (I think the double-quotes after value= are causing that)

rleyva:
Only the most recent build on semaphore contains the fixes for all of these issues. I was aware of this and fixed it.

PR #60, opened 2016-03-05
Missions: Add first iteration mission system

jpanikulam:
- Decided to use txros
- Borrowed heavily from old mission system (Explicitly used old PoseEditor, because it's a goddamn great idea)
- Justification for using txros in the readme
- Added legacy Rise6dof and c3_trajectory_generator as placeholders until replacements are added
- The old kill_handling tool is also in the new legacy folder -- we will make no effort to integrate it in other places, as it will be subsumed by Alarms.

Tests to come, for now we're going to integration test at the mission level. PR'ing this so Tess and I can push through the buoy bumping mission by end of break.


command/sub8_missions/readme.md
```diff
@@ -0,0 +1,112 @@
...
+
+```shell
+    roslaunch sub8_launch waypoint_tool.launch
+```
+
+To get results that are useful, you'll want to set up rviz to visualize tf, /wrench, /posegoal, and /odom (Set odom and posegoal to use axes instead of arrow). I use /odom, keeping a bunch of arrows, for some indication that movement has actually happened.
+
+With the wire facing away from you, right-button will snap the posegoal to the sub's current pose, and left-button will make the sub go to the current posegoal
+
+tags for greppers: 3dmouse 3d-mouse 3dconnexion
```
tessbianchi, line 20: Nice

command/sub8_missions/readme.md
```diff
@@ -0,0 +1,112 @@
...
+
+
+# Running a Mission
+
+```shell
```
tessbianchi, line 27: Make more clear, that these are example missions, and not the process to start a mission.

command/sub8_missions/readme.md
```diff
@@ -0,0 +1,112 @@
...
+        yield sub_singleton.move.height(buoy_location[2]).go()
+        # Wait until we finish matching the buoy's height (We yielded the Deferred)
+        yield sub_singleton.move.set_position(buoy_location).go()
+        # Wait until we match the buoy's position, bumping it
+
+        # back off 2m
+        yield sub_singleton.move.backward(2).go()
+        # Wait until we have backed off
+
+        print "Bumped the buoy"
```
tessbianchi, line 105: Gorgeous

jpanikulam:
Addressed comments.

tessbianchi:
Approved

PR #61, opened 2016-03-05
MAPPER: Make thruster_out alarm faster

jpanikulam:
- and fix a few bugs related to it. 

The new process: instead of updating the bus layout parameter, a list of failed thrusters is cached, and instead of forming a new B matrix every time a thruster fails, we simply adjust the thrust bounds for all of the failed thrusters.


pemami4911:
How significant is the performance increase? 

Looks great, only suggestion is that (we should all) include the status of tests regarding our commits in the PR message so the reviewer has that info readily available.

jpanikulam:
It shaves off a few precious milliseconds that were previously causing issues, and substantially reduces the complexity of a thruster-out alarm

PR #63, opened 2016-03-07
PERCEPTION: Add buoy detection pcl

jpanikulam:
Buoy detection
### What this does
- Simple detection of buoys in 2d, Improvements to come @tessbianchi
- Adds a library of templated helper functions for PCL, and OpenCV/PCL interoperation
  - Adds a bunch of tools for projecting 2d segmentation into 3d
- Adds a tool for visualizing buoy detections in Rviz
- Adds a **very rough** first VisionRequest.srv, that is our first step towards the way missions will request vision information.
- Adds a library called backward.hpp that can be used to debug segfaults, pretty-printing stacktraces (Not worth reviewing, this is from [here.](https://github.com/bombela/backward-cpp))
### What this does not do
- All buoy colors, that's easy from here
- Add the actual buoy bumping mission, this is for another PR

Submitting to David for review.


jpanikulam:
Note that 2,007 of the 2,948 lines in this PR are from backward.hpp, are external, and as such do not need to be reviewed.

gnc/perception/sub8_pcl/include/sub8_pcl/geometry.hpp
```diff
@@ -0,0 +1,133 @@
...
+  PointT point;
+  std::vector<double> distances(cloud.points.size(), 10);  // Initialize to value 10
+// Generate a vector of distances
+#pragma omp parallel for
+  for (size_t index = 0; index < cloud.points.size(); index++) {
+    point = cloud.points[index];
+
+    Eigen::Vector3f cloud_pt(point.x, point.y, point.z);
+    Eigen::Vector3f difference = (line_pt - cloud_pt);
+    double distance = (difference - (difference.dot(direction) * direction)).norm();
```
DSsoto: Shouldnt this be `distance = direction.cross(difference).norm() / direction.norm()` ?
See [http://mathworld.wolfram.com/Point-LineDistance3-Dimensional.html](url), formula (9)
(might be equivalent)

gnc/perception/sub8_pcl/nodes/buoy.cc
```diff
@@ -0,0 +1,291 @@
...
+
+#include "ros/ros.h"
+
+#include "sub8_msgs/VisionRequest.h"
+
+// For stack-tracing on seg-fault
+#define BACKWARD_HAS_BFD 1
+#include <sub8_build_tools/backward.hpp>
+
+#define VISUALIZE
```
DSsoto: Great idea! I'll steal this

gnc/perception/sub8_pcl/nodes/buoy.cc
```diff
@@ -0,0 +1,291 @@
...
+
+// Compute the buoy position in the camera model frame
+// TODO: Check if we have an image
+// TODO: Synchronize cloud and image better
+bool Sub8BuoyDetector::request_buoy_position(sub8_msgs::VisionRequest::Request &req,
+                                             sub8_msgs::VisionRequest::Response &resp) {
+  std::string tf_frame;
+  Eigen::Vector3f position;
+
+  if (!got_cloud) {
```
DSsoto: ROS_ERROR() ?

gnc/perception/sub8_pcl/nodes/buoy.cc
```diff
@@ -0,0 +1,291 @@
...
+    cv::approxPolyDP(contours[i], approx, 5, true);
+    double area = cv::contourArea(approx);
+
+    // Ignore small regions
+    if (area < 100) {
+      continue;
+    }
+    sub::PointXYZT centroid_projected =
+        sub::project_uv_to_cloud(*point_cloud_raw, contour_centroid, camera_model);
+    buoy_centers.push_back(sub::point_to_eigen(centroid_projected));
```
DSsoto: Name could be misleading since we expect these  3D points to be on the surface of the buoy even though they appear in the center in the image. I'm probably nitpicking here.

gnc/perception/sub8_pcl/nodes/buoy.cc
```diff
@@ -0,0 +1,291 @@
...
+  size_t closest_buoy_index = 0;
+
+  for (size_t buoy_index = 0; buoy_index < buoy_centers.size(); buoy_index++) {
+    const double distance = buoy_centers[buoy_index].norm();
+    if (buoy_centers[buoy_index].norm() < closest_distance) {
+      closest_distance = distance;
+      closest_buoy_index = buoy_index;
+    }
+  }
+  center = buoy_centers[closest_buoy_index];
```
DSsoto: Logic: is the closest point necessarily the most likely to belong to the desired buoy? Maybe once you calculate a segmented point cloud you could use a heuristic taking into account the size of the cloud and the average distance and determine if it is indeed "buoy sized" and of the right color.

gnc/perception/sub8_pcl/nodes/buoy.cc
```diff
@@ -0,0 +1,291 @@
...
+}
+
+void Sub8BuoyDetector::image_callback(const sensor_msgs::ImageConstPtr &image_msg,
```
DSsoto: It seems like we are converting images constantly even in the absence of a VisionRequest service call. Maybe we should have a member variable indicating when there's an active request so we don't waste computations.

gnc/perception/sub8_pcl/nodes/buoy.cc
```diff
@@ -0,0 +1,291 @@
...
+    current_image = input_bridge->image;
+  } catch (cv_bridge::Exception &ex) {
+    ROS_ERROR("[draw_frames] Failed to convert image");
+    return;
+  }
+  cam_model.fromCameraInfo(info_msg);
+  image_time = image_msg->header.stamp;
+}
+
+void Sub8BuoyDetector::cloud_callback(const sensor_msgs::PointCloud2::ConstPtr &input_cloud) {
```
DSsoto: Again, it seems like we could skip point cloud processing in the absence of a VisionRequest service call. To deal with image and point cloud synchronization, we could wait until we get a pair with similar enough time-stamps after a request is active and then try to locate the buoy.

gnc/perception/sub8_pcl/include/sub8_pcl/geometry.hpp
jpanikulam: Looks like they're equivalent: [implemented, on wikipedia](https://en.wikipedia.org/wiki/Distance_from_a_point_to_a_line#Vector_formulation), and [wolfram's formula](https://en.wikipedia.org/wiki/Distance_from_a_point_to_a_line#Another_vector_formulation), on wikipedia.

gnc/perception/sub8_pcl/nodes/buoy.cc
jpanikulam: Added ROS_ERROR

DSsoto:
Everything looks pretty good, impressive work! I would have tried to run it but I don't have any buoy bags. 

Since most of our perception will be done using PCL and OpenCV, I suggest maybe having sub8_pcl become sub8_perception or something of the sort. That way we could include the torpedo vision and other vision routines as services or actions along with your buoy stuff and share the infrastructure (especially the pcl code you wrote which would be nice to have available from my end as well). Otherwise the repo will get bloated with pretty much identical copies of a lot of the files in this PR for other perception packages.

Lastly, I was wondering why you didn't make buoy detection an action instead of a service so we could get feedback and stop it prematurely if necessary.

gnc/perception/sub8_pcl/nodes/buoy.cc
jpanikulam: Good catch. I can see this causing problems, if we write missions with the assumption that the returned point is the center of the buoy. I've added a few lines to slide the point along the vector by one buoy radius, to get an approximate center

gnc/perception/sub8_pcl/nodes/buoy.cc
jpanikulam: Good idea. Next version, we should do a combination of distance, deviation in size, and deviation in color.

We also ought to try point cloud geometry matching (RANSAC?) and try to track the thing for a few frames, instead of just the most recent, and shit the bed if we can't track

gnc/perception/sub8_pcl/nodes/buoy.cc
jpanikulam: Fixed this - the image conversion is pretty negligible, but I benchmarked the cloud conversion, which clocked in at 160ms per callback. Moved that into the service call. 

It makes the visualization a little bit laggier, but since the tf cache will make it work in practice.

jpanikulam:
I agree with your suggestion to rename the package, just did that in the final commit (Might have broken the build, bear with me).

I didn't make it an action because there's not much intermediate feedback, really we either detect the buoy and get a centroid, or we get no results at all. I suspect most vision tasks would involve repeated calls to a service instead of an action. Maybe actions will be more appealing after we do testing.

Stand-by for commits + merging master in

jpanikulam:
Addressed comments in [this commit](https://github.com/jpanikulam/Sub8/commit/c22bbc4ba3e39c7cbbac473db968c0a5c82c6b79), and also renamed the package. Will merge pending final review.

Great comments, btw

PR #67, opened 2016-03-16
ANYTIME_PLANNER: Added OMPL's Anytime Planner

pemami4911:
Main additions: 
- Anytime Planner
- Easy addition at run-time of geometric obstacles for testing 

TGEN status: Ready to be integrated with the profiler


jpanikulam:
Build fail! Otherwise, I see no glaring problems (Looks like a file is missing)

pemami4911:
I think the version of OMPL that Semaphore is using doesn't have this file, according to [http://ompl.kavrakilab.org/releaseNotes.html](url) it was added in version 1.0.0... Investigating further

jpanikulam:
Hrm. Semaphore grabs [libompl-dev](http://packages.ubuntu.com/trusty/libompl-dev), which is the same version most subbians arebuilding with. Looks super out of date. Do we need to all build from source?

@whispercoros (Anthony) -- we really need that ppa, or every person will burn hours building things from source on first install.

sentree:
I can build the PPA, I will just need a machine to use in the lab. Is the one storing all of the ros bags just for storage, or is something running on it? If not, I can probably just use that.

jpanikulam:
@whispercoros Go ahead and use the bag machine, it has nothing but lonely bags and lots of hdd space.

jpanikulam:
@whispercoros -- Don't prioritize ppa over ieee, but how long do you think until we can close this?

sentree:
@jpanikulam -- Been very busy with IEEE as you know; however, in order to set up something that we can actually pull from, we need a way to access it. There are two options available:
1. We get a wireless access point for the server so that the PPA can be hosted on a MIL network of sorts (for in the lab only).
2. We find some server with a static IP and use that as a frontend to pull from via some crazy networking stuff (could be used anywhere).

Of course we could always do both if access from both is needed. Currently, I am waiting on a NIC from Daniel to implement the first option. That should be in sometime next week. If you want something sooner, get me a static IP somewhere and I can get it running.

jpanikulam:
@pemami4911 ded?

pemami4911:
Did the PPA happen so that the build issues with this get resolved? 

(What is dead may never die)

rleyva:
No, the repo hasn't happened yet. Do you by any chance know if the MoveIt! library has a more updated OMPL? If it does we can just pull that from the Ubuntu repos and sidestep this for now.

sentree:
I have everything in place for a local PPA, I would just need the compiled binaries or a link to somewhere with the source so I could compile and package them. I am still waiting on the static IP address from Daniel for an internet facing PPA. Hope MoveIt! works out though, it would be much simpler to implement.

DSsoto:
Closing to have clean PR slate for competition. We should get that PPA or MoveIt lib and add this after the competition. @rleyva, @whispercoros any updates on this?

sentree:
I decided to wait to PR the scripts because the competition was so close. As soon as that is done, this is my next order of business (specifically, see issue #89).

PR #68, opened 2016-03-22
PERCEPTION: add automatic robust image segmentation

DSsoto:
This pull request adds robust segmentation capabilities to the sub8_perception package.
It is designed to segment out foreground objects of solid color whose color distribution does not significantly overlap with that of the background. It tries to find the foreground distribution's mode and calculate ideal threshold values by finding specific zero_crossings of the histogram's derivative. 

This method makes segmentation of solid colors extremely robust to lighting changes and some color loss, since no hardcoded thresholds are used. In my tests, telling it to look for hue distribution peaks from anywhere from red to green all resulted in successful segmentation of the yellow torpedo board. 

The following image shows this in action: 
screenshot from 2016-03-22 04 53 48

Assigning to @rleyva  and @jpanikulam for review.


gnc/sub8_perception/include/sub8_pcl/cv_tools.hpp
```diff
@@ -20,6 +26,28 @@ typedef std::vector<cv::Point> Contour;
...
+std::vector<cv::Point> find_local_maxima(const cv::MatND &histogram,  float thresh_multiplier);
+
+// Selects the mode of a multi-modal distribution closest to a given target value
+unsigned int select_hist_mode(std::vector<cv::Point> &histogram_modes, unsigned int target);
+
+// Takes in a grayscale image and segments out a semi-homogenous foreground object with
+// pixel intensities close to <target>. Tuning of last three parameters may imrove
+// results but default values should work well in most cases.
+void statistical_image_segmentation(const cv::Mat &src, cv::Mat &dest, 
+		const int hist_size, const float** ranges, const int target, std::string image_name,
```
jpanikulam, line 50: Minor formatting - don't bother fixing, it will be auto-fixed the next time somebody clang-formats

gnc/sub8_perception/nodes/buoy.cc
```diff
@@ -319,8 +341,99 @@ sub::statistical_outlier_filter<sub::PointXYZT>(scene_buffer, current_cloud,
...
+  cv::resize(segmented_board, segmented_board, cv::Size(0,0), 2, 2);
+  return segmented_board;
+}
+
+
 int main(int argc, char **argv) {
   ros::init(argc, argv, "pcl_slam");
-  boost::shared_ptr<Sub8BuoyDetector> sub8_buoys(new Sub8BuoyDetector());
+  ROS_INFO("Initializing node /pcl_slam");
+  // boost::shared_ptr<Sub8BuoyDetector> sub8_buoys(new Sub8BuoyDetector());
```
jpanikulam: Probably meant to uncomment?

Separate classes is a good idea. Next iteration, let's start uncluttering buoy.cc

gnc/sub8_perception/src/sub8_pcl/cv_utils.cc
```diff
@@ -1,12 +1,270 @@
...
+	for (int i = 0; i < offset; ++i)  // Pad filtered result with zeroes
+	{
+		hist.at<float>(i) = 0;
+		hist.at<float>(histSize - 1 - i) = 0;
+	}
+	return hist;
+}
+
+
+std::vector<float> generate_gaussian_kernel_1D(size_t kernel_size, float sigma){
```
jpanikulam, line 35: Nice. [Boost has an implementation of this](http://www.boost.org/doc/libs/1_53_0/libs/math/doc/sf_and_dist/html/math_toolkit/dist/dist_ref/dists/normal_dist.html)

You can use it like [this](https://github.com/jpanikulam/Sub8/blob/slam/gnc/sub8_slam/analysis/bayesian_filter.cc#L64)

and get the pdf like [this](https://github.com/jpanikulam/Sub8/blob/slam/gnc/sub8_slam/analysis/bayesian_filter.cc#L66)

jpanikulam, line 35: Based on [their implementation](http://www.boost.org/doc/libs/master/boost/math/distributions/normal.hpp) yours might be faster because it assumes 0 mean and skips some frivolous checks.

gnc/sub8_perception/src/sub8_pcl/cv_utils.cc
```diff
@@ -1,12 +1,270 @@
...
+}
+
+
+std::vector<cv::Point> find_local_maxima(const cv::MatND &histogram, float thresh_multiplier){
+	std::vector<cv::Point> local_maxima, threshed_local_maxima;
+	float global_maximum = 0;
+
+	// Locate local maxima and find global maximum
+	for(size_t idx = 1; idx < histogram.total() - 1; idx++){
+		float current_value = histogram.at<float>(idx);
```
jpanikulam, line 61: confused tabbing

gnc/sub8_perception/src/sub8_pcl/cv_utils.cc
```diff
@@ -1,12 +1,270 @@
...
+	    if((histogram.at<float>(idx-1) < current_value) && (histogram.at<float>(idx+1) <= current_value)){
```
jpanikulam, line 62: No issue since speed is not  concern, but you could save some cycles if you cache previously accessed indices. I don't think opencv mats are structured in a way that lets the compiler figure this out.

gnc/sub8_perception/src/sub8_pcl/cv_utils.cc
```diff
@@ -1,12 +1,270 @@
...
+	    	local_maxima.push_back(cv::Point(idx, current_value));
+	    	if(global_maximum < current_value) global_maximum = current_value;
+	    }
+	}
+	// std::cout << std::endl << "Maxima: ";
+	BOOST_FOREACH(cv::Point pt, local_maxima){
+		if(pt.y > global_maximum * thresh_multiplier) threshed_local_maxima.push_back(pt);
+		// std::cout << boost::format("[%1%, %2%] ") % pt.x % pt.y;
+	}
+	// std::cout << std::endl << boost::format("thresh: > global_maximum(%1%) * thresh_multiplier(%2%) = %3%")
```
jpanikulam: Delete dead code

gnc/sub8_perception/src/sub8_pcl/cv_utils.cc
```diff
@@ -1,12 +1,270 @@
...
+	}
+	// Normalize kernel (sum of values should equal 1.0)
+	float sum = 0;
+	for (size_t i = 0; i < kernel_size; i++){ sum += kernel[i];	}
+	for (size_t i = 0; i < kernel_size; i++){ kernel[i] /= sum;	}
+	return kernel;
+}
+
+
+std::vector<cv::Point> find_local_maxima(const cv::MatND &histogram, float thresh_multiplier){
```
jpanikulam, line 55: We need to start compiling with C++11 so we can take advantage of move semantics. Otherwise, this does a copy

@whispercoros Counting on you for that PPA!

gnc/sub8_perception/src/sub8_pcl/cv_utils.cc
```diff
@@ -1,12 +1,270 @@
...
+	cv::Rect upper_right_corner = cv::Rect(histImage.cols - dest.cols - 1, 0, dest.cols, dest.rows);
+	dest.copyTo(segmentation_channel(upper_right_corner));
+	segmentation_channel = segmentation_channel * 0.3;
+	std::vector<cv::Mat> debug_img_channels;
+	debug_img_channels.push_back(histImage); 
+	debug_img_channels.push_back(histDerivImage);
+	debug_img_channels.push_back(segmentation_channel);
+	cv::Mat debug_img;
+	cv::merge(debug_img_channels, debug_img);
+	cv::imshow((boost::format("Statistical Image Segmentation(%1%)") % image_name).str(), debug_img );
```
jpanikulam, line 296: remove debug imshow & misc debug stuff in library function

gnc/sub8_perception/src/sub8_pcl/cv_utils.cc
```diff
@@ -1,12 +1,270 @@
...
+	hist_derivative.at<float>(0) = 0;
+	hist_derivative.at<float>(hist_size - 1) = 0;
+	for (int i = 1; i < hist_size - 1; ++i)
+	{
+		hist_derivative.at<float>(i) = (hist_smooth.at<float>(i + 1) - hist_smooth.at<float>(i - 1)) / 2.0;
+	}
+	hist_derivative = sub::smooth_histogram(hist_derivative, kernel_size, sigma);
+
+	// Find target mode
+	// std::cout << boost::format("Target: %1%") % target;
```
jpanikulam: kill commented debugs

gnc/sub8_perception/src/sub8_pcl/cv_utils.cc
```diff
@@ -1,12 +1,270 @@
...
+	// 		std::cout << boost::format(" %1% ") % pt.x;
+	// 	}
+	// }
+	// else std::cout << "same as 'Maxima'";
+	// std::cout << std::endl;
+	return threshed_local_maxima;
+}
+
+
+std::vector<cv::Point> find_local_minima(const cv::MatND &histogram, float thresh_multiplier){
```
jpanikulam, line 92: Add brief doc -- what exactly does thresh_multiplier do?

gnc/sub8_perception/src/sub8_pcl/cv_utils.cc
```diff
@@ -1,12 +1,270 @@
...
+		% target % target_mode % low_thresh % high_thresh ).str() ;
+	ROS_INFO(ros_log.c_str());
+
+	// Threshold image
+	cv::inRange(src, low_thresh, high_thresh, dest);
+
+	#ifdef VISUALIZE
+	// Prepare to draw graph of histogram and derivative
+	int hist_w = 512; int hist_h = 400;
+	int bin_w = cvRound( (double) hist_w/hist_size );
```
jpanikulam, line 259: afaik this is equivalent to hist_w / hist_size, no?

gnc/sub8_perception/nodes/buoy.cc
```diff
@@ -319,8 +341,99 @@ sub::statistical_outlier_filter<sub::PointXYZT>(scene_buffer, current_cloud,
...
+  ROS_ERROR("Error constructing Sub8TorpedoBoardDetector using initializer list: ");
+  ROS_ERROR(e.what());
+}
+
+
+Sub8TorpedoBoardDetector::~Sub8TorpedoBoardDetector(){
+
+}
+
+cv::Mat torpedo_board_segmentation(const cv::Mat &image);
```
jpanikulam: who is this guy?

gnc/sub8_perception/nodes/buoy.cc
```diff
@@ -319,8 +341,99 @@ sub::statistical_outlier_filter<sub::PointXYZT>(scene_buffer, current_cloud,
...
+  sub::statistical_image_segmentation(sat_blurred, threshed_sat, hist_size, ranges,
+                                      target_saturation, "Saturation", 6, 0.1, 0.1);
+  segmented_board = threshed_hue / 2.0 + threshed_sat / 2.0;
+  cv::threshold(segmented_board, segmented_board, 255*0.75, 255, cv::THRESH_BINARY);
+  
+#ifdef VISUALIZE
+  cv::imshow("segmented board", segmented_board);
+#endif
+
+  cv::resize(segmented_board, segmented_board, cv::Size(0,0), 2, 2);
```
jpanikulam, line 424: What does this do? cv::Size(0, 0)?

gnc/sub8_perception/nodes/buoy.cc
```diff
@@ -319,8 +341,99 @@ sub::statistical_outlier_filter<sub::PointXYZT>(scene_buffer, current_cloud,
...
+}
+
+
+bool Sub8TorpedoBoardDetector::request_torpedo_board_position(sub8_msgs::VisionRequest::Request &req,
+                                           sub8_msgs::VisionRequest::Response &resp){
+  return true;
+}
+
+
+cv::Mat Sub8TorpedoBoardDetector::board_segmentation(const cv::Mat &image){
```
jpanikulam: Pass large outputs by reference -- cv::Mat doesn't have move semantics

gnc/sub8_perception/src/sub8_pcl/cv_utils.cc
DSsoto, line 92: The doc is in cv_tools.hpp. Do you still want something here?

gnc/sub8_perception/nodes/buoy.cc
DSsoto: removed

gnc/sub8_perception/src/sub8_pcl/cv_utils.cc
```diff
@@ -1,12 +1,270 @@
...
+	float sum = 0;
+	for (size_t i = 0; i < kernel_size; i++){ sum += kernel[i];	}
+	for (size_t i = 0; i < kernel_size; i++){ kernel[i] /= sum;	}
+	return kernel;
+}
+
+
+std::vector<cv::Point> find_local_maxima(const cv::MatND &histogram, float thresh_multiplier){
+	std::vector<cv::Point> local_maxima, threshed_local_maxima;
+	float global_maximum = 0;
```
jpanikulam: Does this, and the corresponding line in find_minima, assume that the input is strictly positive? 

Definitely true for the histogram, but not guaranteed for histogram derivative.

[stl has an implementation of infinity](http://en.cppreference.com/w/cpp/types/numeric_limits/infinity)

jpanikulam:
This is great code, David.

Mostly minor formatting and a few questions. If you're using sublime, I'd suggest installing the [Sublime clang-format plugin](https://github.com/rosshemsley/SublimeClangFormat) so you don't have to fuss with whitespace.

Will give @rleyva  a change to take a look as well.

gnc/sub8_perception/src/sub8_pcl/cv_utils.cc
DSsoto, line 259: Counterexample: 50.0/26.0 = 1.9231 which gets rounded up to 2.
50 / 26 = 1 (integer division)

gnc/sub8_perception/src/sub8_pcl/cv_utils.cc
jpanikulam, line 92: Nope. My bad.

gnc/sub8_perception/nodes/buoy.cc
DSsoto, line 424: Tells resize to ignore size argument and just base itself off of the x and y scaling factors (both 2 in this case)

gnc/sub8_perception/src/sub8_pcl/cv_utils.cc
jpanikulam, line 259: gotcha

gnc/sub8_perception/nodes/buoy.cc
jpanikulam, line 424: gotcha

gnc/sub8_perception/src/sub8_pcl/cv_utils.cc
DSsoto: The only reservation I have is that these commented debug statements will likely prove very useful for quick tuning at the competition. What do you say?

jpanikulam: Perhaps conceal them in a #ifdef VISUALIZE so it's clear that it's not dead code

gnc/sub8_perception/src/sub8_pcl/cv_utils.cc
DSsoto, line 296: Even though it's surrounded by "#ifdef VISUALIZE  ...  #endif" ?

jpanikulam, line 296: Oh - that's fine, I missed the ifdef. Preprocessor directives should be the start of the line, so the if/endif is clear

gnc/sub8_perception/src/sub8_pcl/cv_utils.cc
DSsoto, line 61: It looks normal on my machine. Weird.
Deleted whitespace and re-tabbed

gnc/sub8_perception/src/sub8_pcl/cv_utils.cc
DSsoto, line 35: Ok, I'll leave it as is then.

gnc/sub8_perception/nodes/buoy.cc
DSsoto: Yea, that was supposed to be temporarily commented out only.

gnc/sub8_perception/include/sub8_pcl/cv_tools.hpp
```diff
@@ -13,6 +17,8 @@
 #include <sensor_msgs/image_encodings.h>
 #include "ros/ros.h"
 
+#define VISUALIZE
```
jpanikulam, line 21: This shadows the #define VISUALIZE in buoy.cc and will propagate through includes. Let's remove the one in buoy.cc

gnc/sub8_perception/nodes/buoy.cc
DSsoto: Maybe put our main in perception.cc and separate our class implementations into buoy.cc, buoy.hpp, torpedo_board.cc and torpedo_board.hpp? What do you think? 

I also don't really like the fact that the header for cv_utils.cc is called  cv_tools.hpp.

gnc/sub8_perception/src/sub8_pcl/cv_utils.cc
DSsoto, line 62: It's a one dimensional mat which is probably easy and quick to access. If you think it'll make a difference, I could copy its values into a vector and then work off of that.

gnc/sub8_perception/src/sub8_pcl/cv_utils.cc
DSsoto, line 55: I agree, so should any changes be made here?

gnc/sub8_perception/nodes/buoy.cc
jpanikulam: cv_tools is intended to be the header for more than one cv_*.cc, much like pcl_tools. Originally, cv_utils had just a handful of miscellaneous utility functions

gnc/sub8_perception/include/sub8_pcl/cv_tools.hpp
DSsoto, line 21: removed

gnc/sub8_perception/src/sub8_pcl/cv_utils.cc
jpanikulam, line 55: Since we intend to switch anyway, I'd say no point

gnc/sub8_perception/src/sub8_pcl/cv_utils.cc
DSsoto, line 296: Yup, my bad, fixed it.

gnc/sub8_perception/src/sub8_pcl/cv_utils.cc
DSsoto: Not making that assumption, changed it to +/- std::numeric_limits<double>::infinity()

gnc/sub8_perception/src/sub8_pcl/cv_utils.cc
DSsoto: concealed them in #ifdef SEGMENTATION_DEBUG

gnc/sub8_perception/nodes/buoy.cc
DSsoto: Passing by reference now

DSsoto:
@jpanikulam Thanks, I'll make sure to install that, whitespaces were weird and inconsistent in Sublime. Pushed new commit addressing your comments.

jpanikulam:
Merged.

Inconsistent whitespacing is probably because of mixed tabs and spaces. ctrl+a in subl will show you whitespace types

PR #70, opened 2016-03-23
POOL: Add pool day modifications

jpanikulam:
These are the changes that were made on the pool day at Jake Easterling's apartment.


jpanikulam:
Reassigning from @rleyva so that E-Fields doesn't ruin his life

pemami4911:
Looks fine! Read the mission planner wiki; tx_ros is cool :+1:

PR #73, opened 2016-03-27
VISION: Improve buoy detection, visualization flags

jpanikulam:
- VISUALIZE is no longer in any headers, instead, do `catkin_make -DVISUALIZE=true`
  - To remake without visualize, do `catkin_make -DVISUALIZE=false`
  - This is to guarantee that we never ever accidentally compile on the sub with visualization enabled
    - (We'd get all kinds of silly x-server crashes)
  - And so that you don't have to fish around the .hpp's to find the def
- Put vision geometry parameters (Like buoy radius) in a yaml in sub8_launch/config
- Put vision color calibrations (Like buoy/red color) in a yaml in sub8_launch/config
- Made buoy detection more robust
  - In the past, if the point cloud was missing at the buoy, we'd accidentally "project" to the nearest point, which was something random
  - Now, we only accept the projection if when projected back onto the 2d image, it falls nearly within the original buoy contour
- clang-formatted cv_utils


gnc/sub8_perception/src/sub8_pcl/buoy.cc
```diff
@@ -171,7 +218,10 @@ void Sub8BuoyDetector::determine_buoy_position(
       closest_buoy_index = buoy_index;
     }
   }
+
   center = buoy_centers[closest_buoy_index];
+
+  return true;
 }
 
 void Sub8BuoyDetector::image_callback(const sensor_msgs::ImageConstPtr &image_msg,
```
DSsoto, line 241: I know this isnt from this PR but I'd suggest that you just update pointers to the image_msg and info_msg here and only do the cv_bridge stuff inside your service callback function. That way we're only decoding images when they actually need to be processed.

gnc/sub8_perception/src/sub8_pcl/buoy.cc
```diff
@@ -210,19 +260,14 @@ void Sub8BuoyDetector::cloud_callback(const sensor_msgs::PointCloud2::ConstPtr &
...
-                                                );
-                                                */
 #ifdef VISUALIZE
+  sub8_msgs::VisionRequest::Request req;
+  req.target_name = "red";
+  sub8_msgs::VisionRequest::Response resp;
+  request_buoy_position(req, resp);
   pcl::console::print_highlight("Getting Point Cloud\n");
   if (!got_cloud) {
     pcl::console::print_highlight("Getting new\n");
```
DSsoto: Getting new what?

gnc/sub8_perception/src/sub8_pcl/cv_param_helpers.cc
```diff
@@ -0,0 +1,30 @@
...
+
+namespace sub {
+
+/// Range is an output
+void range_from_param(std::string &param_root, Range &range) {
+  /// Ex: param_name = /buoy/red --
+
+  // No overhead in doing this
+  ros::NodeHandle nh;
+  ROS_WARN("-- Waiting for parameter %s, (No timeout)", param_root.c_str());
```
DSsoto: Seems like this one and the next should be using ROS_INFO. 

> WARN
> 
> Information that the user may find alarming, and may affect the output of the application, but is part of the expected working of the system. Examples:
> "Could not load configuration file from <path>. Using defaults."

gnc/sub8_perception/src/sub8_pcl/cv_param_helpers.cc
```diff
@@ -0,0 +1,30 @@
+#include <sub8_pcl/cv_tools.hpp>
+#include <ros/console.h>
+
+namespace sub {
+
+/// Range is an output
+void range_from_param(std::string &param_root, Range &range) {
```
DSsoto, line 7: Is Range a class that we defined?

DSsoto:
Minor comments, looks good!
I like that you're standarizing color ranges through that yaml file.

gnc/sub8_perception/src/sub8_pcl/cv_param_helpers.cc
jpanikulam: Gotcha. I think this is closer to a warning, because the whole buoy detector will hang on this until the param exists. The user should be warned that the node might stop operating

gnc/sub8_perception/src/sub8_pcl/cv_param_helpers.cc
jpanikulam, line 7: yea, in cv_tools.hpp

gnc/sub8_perception/src/sub8_pcl/buoy.cc
jpanikulam, line 241: Changed

PR #76, opened 2016-03-28
CAMERA: Add stereo launch file

rleyva:


jpanikulam:
Thanks a billion, Ralph

Can you move the stereo launches (The ones you implemented yourself) to sub8_launch, and include this in the forward cameras launch file (Or wherever you see it fitting best)?

rleyva:
The stereo node will take the CPU usage way up once the cameras come online. If we're not running cameras or are not interested in point cloud data for the time being we'll be using up battery for no reason. Shouldn't we bring up stereo when we need it? It's your call.

jpanikulam:
I see, v. good point. In that case, still move the launch files, and then <include> both this and the camera launches in sub8.launch

jpanikulam:
Nevermind, will do right now

PR #78, opened 2016-03-28
VISION: Add service for channel markers

jpanikulam:
Small additions to Brandon's code, to work with the mission system


Peach8:
Approved

PR #80, opened 2016-03-29
gazebo package & a launch file

guojiyao:
That's the gazebo pkgs and a test launch file needed for the simulator. Will also write a .sh file to install gazebo later.


jpanikulam:
Included by #81

Issue #77, opened 2016-03-28
Decentralize alarm system


jpanikulam:
- Easier sharing between robots
- Easier implementation of alarm handling, instead of a sort of lanky and cumbersome service-based organization
- This requires no major changes to the alarm node

jpanikulam:
Customer: @zachgoins

PR #94, opened 2016-04-09
ALARMS: Added an alarm listener class.

mattlangford:


command/sub8_alarm/sub8_alarm/alarm_helpers.py
```diff
@@ -91,3 +91,50 @@ def raise_alarm(self, problem_description=None, parameters=None, severity=None):
...
+            return
+
+        callback = found_alarm['callback']
+        args = found_alarm['args']
+
+        #Actaully call the callback function and pass args if nessicary.
+        if args is None:
+            callback(alarm)
+        else:
+            callback(alarm, args)
```
jpanikulam, line 140: Use *args so the args can be passed directly

jpanikulam:
Matt: great stuff, I love the comments. 
- In the future, if you pull --rebase you can avoid having to merge. But sometimes you have to merge and that's ok. 
- run flake8 --max-line-length=130 on this file, there are some minor violations

jpanikulam:
This closes #77 
@zachgoins

command/sub8_alarm/sub8_alarm/alarm_helpers.py
mattlangford, line 140: Just to confirm before I recommit - if there are multiple args, they still have to be passed to the listener as a list but they will be sent out as multiple arguments to the function callback.

zachgoins:
@mattlangford @jpanikulam you guys da best

command/sub8_alarm/sub8_alarm/alarm_helpers.py
jpanikulam, line 140: Yep.
 If you have add_listener take a *args, then you can pass a bunch of arguments in the same way that you send them to the callback.

jpanikulam:
Merging now so Zach can use - address minor stuff in a future commit

forrestv:
Seven minutes too late, but:

There's really no reason to have an explicit `args`/`*args` argument. I know Twisted has a lot of that kind of thing in their API, but I think that that's just because Twisted started before the `lambda` keyword was introduced to Python.

With `lambda`, you can create your own little anonymous function to add the extra arguments if you need to, without adding complexity to the API for everyone who doesn't need to. To the user it'd be something like:

```
alarm_listener.add_listener(alarm_name, lambda alarm: my_real_callback(alarm, 1, 2, 3))
```

rather than

```
alarm_listener.add_listener(alarm_name, my_real_callback, 1, 2, 3)
```

This simplifies AlarmListener a lot too.

mattlangford:
That's an interesting way to do it. I didn't know you could pass lamdas around like that though I shouldn't be surprised. I was modeling the callback like a ros subscriber where you can pass optional arguments in, but this definitely simplifies it a lot - I will make this change for my next commit.

PR #95, opened 2016-04-11
VISION: Seek channel markers using NCC

jpanikulam:
- Lean heavily on normalized cross-correlation to detect channel markers, lessening dependence on color
- Also includes updates to Gazebo to support more accurate channel marker chasing

**How it does**
Visualization


simulation/sub8_gazebo/nodes/gazebo_controller.py
```diff
@@ -27,13 +35,34 @@ def send_wrench(self, wrench):
...
+            )
+        ))
+        return ResetGazeboResponse()
+
     def wrench_cb(self, msg):
         wrench = msg.wrench
         self.send_wrench(wrench)
 
+    def publish_height(self, z):
+        '''HACK. Lanford to fix'''
```
mattlangford: Langford\* 
No big deal.

mattlangford:
Are there scenarios when this method would fail (maybe if there's a shadow on the marker or if the bottom of the transdec is particularly reflective)?

simulation/sub8_gazebo/nodes/gazebo_controller.py
jpanikulam: OOPS FIXED

jpanikulam:
Fair question - It substantially improved reliability over color alone. This change made us perform much better around noise, and when the thing was half occluded (the tether went over it), it still behaved well, gave a good heading

That said, the right solution is an SVM or a very small neural net trained on {r, g, b, ncc} instead of the hackish and-ing. But we need to spend time labelling data for that to be a worthy enterprise!

PR #104, opened 2016-04-20
Initial Commit for ActuatorBoard

duggerdufl:
Added readme and program C file.


jpanikulam:
Great stuff, +1 sub points for solid documentation. 

*sub points redeemable only at participating retailers

khaledhassan:
Late-ish to the party but you may consider including a .bin/.hex for the firmware in the git repo. Generally it's bad mojo to include build artifacts in the repo, but this one currently depends on the mikroC toolchain which is proprietary. Having a "known good" firmware blob would let you more easily replace the board if anything bad happened to it like last time.

PR #106, opened 2016-04-22
ActuatorBoard: Added FW hex, adjusted comments

duggerdufl:
Added hex firmware file for future boards, adjusted comments in C file
and added clarifications to readme


jpanikulam:
@khaledhassan

PR #107, opened 2016-04-22
ACTUATOR: Working with a response from board.

mattlangford:
The marker_occ_grid.py file needs to be tested with data from sub simulations. Right now it uses a simulated camera with simulated camera_info but this will be fixed shortly.

### DIFF
--- a/drivers/sub8_actuator_driver/scripts/actuator_driver.py
+++ b/drivers/sub8_actuator_driver/scripts/actuator_driver.py
@@ -27,7 +27,7 @@ class ActuatorDriver():
     TODO: Wait for response to return True or False. The board does not actually send
         back responses right now. Needs to handle physical disconnect issues.
     '''
-    def __init__(self, port, baud=9600):
+    def __init__(self, port, baud=115200):
         self.load_yaml()
 
         rospy.init_node("actuator_driver")
@@ -45,6 +45,7 @@ class ActuatorDriver():
         )
 
         self.ser = serial.Serial(port=port, baudrate=baud, timeout=None)
+        self.ser.flushInput()
 
         # Reset all valves
         rospy.loginfo("Resetting valves.")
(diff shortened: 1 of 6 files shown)


gnc/sub8_perception/nodes/marker_occ_grid.py
```diff
@@ -0,0 +1,160 @@
...
+
+        # Calculate angle between vectors and use that to find r
+        theta = np.arccos(np.dot(mid_ray, edge_ray))
+        return np.tan(theta) * depth
+
+    def check_for_trench(self):
+        self.trench_service('trench')
+
+
+def projectPixelTo3dRay(uv, temp, depth):
```
jpanikulam, line 147: Temp?

install.sh
```diff
@@ -44,7 +44,7 @@ sudo sh -c 'echo "deb-src http://archive.ubuntu.com/ubuntu trusty main universe"
 sudo sh -c 'echo "deb http://packages.ros.org/ros/ubuntu $(lsb_release -sc) main" > /etc/apt/sources.list.d/ros-latest.list'
 sudo sh -c 'echo "deb http://packages.osrfoundation.org/gazebo/ubuntu trusty main" > /etc/apt/sources.list.d/gazebo-latest.list'
 
-sudo sh -c 'echo "nameserver 8.8.8.8" > /etc/resolv.conf'
+# sudo sh -c 'echo "nameserver 8.8.8.8" > /etc/resolv.conf'
```
jpanikulam, line 47: Very good

gnc/sub8_perception/nodes/marker_occ_grid.py
```diff
@@ -0,0 +1,160 @@
...
+
+import cv2
+import numpy as np
+
+
+def unit_vector(vect):
+    return vect / np.linalg.norm(vect)
+
+
+def make_2D_rotation(angle):
```
jpanikulam, line 19: Check out [this](https://github.com/uf-mil/Sub8/blob/master/utils/sub8_ros_tools/sub8_ros_tools/geometry_helpers.py)

Don't worry about it now, but if you find yourself needing this function elsewhere, put it there

gnc/sub8_perception/nodes/marker_occ_grid.py
mattlangford, line 147: This was just to return different vectors for each call of the function since the image geometry class was not yet used. When the real image and camera_info exist in the script, this temp function will be removed and the function calls will be modified.

drivers/sub8_actuator_driver/scripts/actuator_driver.py
```diff
@@ -59,18 +59,19 @@ def __init__(self, port, baud=9600):
...
-        #             problem_description='The board appears to be disconnected.'
-        #         )
-        #         rospy.sleep(5)
+        #rospy.spin()
+        # Might comment this whole deal out once it all works for sure.
+        r = rospy.Rate(.2)  # hz
+        while not rospy.is_shutdown():
+            # Heartbeat to make sure the board is still connected.
+            r.sleep()
+            self.ping()
```
jpanikulam, line 69: Matt - it appears to be possible to accidentally send a ping command concurrently with one of the service calls (they are called in seperate threads). I would suggest thread-locking using sub8_ros_tools.threading_helpers

@duggerd noticed this

mattlangford, line 69: Yep, ran into this problem with Congregator comms where spamming data could cause packet loss. I was going to just not ping and use service calls as pings but locking the thread will be much much better. Will implement tomorrow and recommit with that to this PR.

mattlangford, line 69: Just recommitted

jpanikulam:
Approved, pulling on build.

PR #114, opened 2016-04-25
MARKER: Add mission to go to and align with marker.

mattlangford:
Seemed to work when generating target poses but controller simulation errors would mess things up.

### DIFF
--- a/command/sub8_missions/missions/align_channel.py
+++ b/command/sub8_missions/missions/align_channel.py
@@ -1,49 +1,72 @@
 from txros import util
+import rospy
+import tf
 from sub8_ros_tools import normalize, clip_norm
+from image_geometry import PinholeCameraModel
 import numpy as np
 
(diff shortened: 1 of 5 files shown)


command/sub8_missions/missions/align_channel.py
```diff
@@ -1,49 +1,72 @@
...
 @util.cancellableInlineCallbacks
-def get_above(sub):
-    '''TODO:
-    Use knowledge of height from DVL
+def run(sub_singleton):
+    # Right now we assume the marker is in our field of view. (Will add searching later)
+    # Go to max height to see as much as we can.
+
+    rospy.init_node('align_channel_mission')
+    tf_listener = tf.TransformListener()
```
jpanikulam, line 15: Later, use the sub object's tf listener so that you don't have to wait for a buffer

PR #119, opened 2016-04-30
SIM: Add job_runner to test jobs.

mattlangford:
In it's current implementation this PR requires `get_last_message_time` from txros which is not currently in the txros master branch.

Also, I think David's PR wrote over some changes that I had made in an earlier push. The final few commits in this PR should fix those changes.


simulation/sub8_gazebo/diagnostics/set_model_position_test.py
```diff
@@ -0,0 +1,64 @@
...
+import tf
+from sub8_ros_tools import msg_helpers
+from nodes import job_runner
+
+from twisted.internet import reactor, defer
+import numpy as np
+
+
+@txros.util.cancellableInlineCallbacks
+def run_mission(srv, sub, nh):
```
mattlangford, line 18: It's funny, this actually tests to make sure the `set_model_position` function works.
For some reason the sub can not maintain arbitrary quaternion poses in the simulator, so I had to limit to only yawing, which worked 100/100 times.

simulation/sub8_gazebo/nodes/job_runner.py
```diff
@@ -0,0 +1,264 @@
...
+            twist.linear.z = 0
+
+        model_state = ModelState()
+        model_state.model_name = model
+        model_state.pose = pose
+        model_state.twist = twist
+
+        if model == 'sub8':
+            kill = nh.get_service_client('/set_kill', SetKill)
+            yield kill(SetKillRequest(kill=Kill(active=True)))
```
mattlangford, line 245: This seems like a jank way to reset the sub's position but it worked. Let me know if you think this should be changed.

jpanikulam, line 245: Love it
- Jake

jpanikulam:
Looks like Semaphore make is failing due to a missing srv file?

mattlangford:
I had an issue with a few missing service files but I thought I had added them back before pushing, will do that tomorrow.

mattlangford:
Looks like the Semaphore failed -- I think it's a build issue.

PR #120, opened 2016-05-01
VISION: Add particle filter for multi-view pose estimation

jpanikulam:
Fun for the whole family, the new Particle Filter from [Cinco](https://en.wikipedia.org/wiki/List_of_Cinco_Family_products) can estimate the pose of anything you want, from roughly spherical things to exactly spherical things!
- Adds some rviz python stuff
- Logs .dae files as binary (even though they are human-readable, they aren't usefully human readable)
- Add last_image_time, as the stamp of the last image gotten to Image_Subscriber
- Add the particle filter itself

### DIFF
--- a/utils/sub8_ros_tools/sub8_ros_tools/image_helpers.py
+++ b/utils/sub8_ros_tools/sub8_ros_tools/image_helpers.py
@@ -66,8 +66,13 @@ class Image_Subscriber(object):
         Assumes topic of type "sensor_msgs/Image"
         This behaves like a conventional subscriber, except handling the additional image conversion
         '''
+        if callback is None:
+            callback = lambda im: setattr(self, 'last_image', im)
+
         self.encoding = encoding
         self.camera_info = None
+        self.last_image_time = None
+        self.last_image = None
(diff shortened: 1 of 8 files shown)


utils/sub8_ros_tools/sub8_ros_tools/image_helpers.py
```diff
@@ -66,8 +66,13 @@ def __init__(self, topic, callback=None, encoding="bgr8", queue_size=1):
         Assumes topic of type "sensor_msgs/Image"
         This behaves like a conventional subscriber, except handling the additional image conversion
         '''
+        if callback is None:
+            callback = lambda im: setattr(self, 'last_image', im)
+
         self.encoding = encoding
         self.camera_info = None
+        self.last_image_time = None
```
mattlangford, line 74: It'd be a good idea to refactor `follow_orange_pipes.py` to use this timestamp instead of its own.

jpanikulam, line 74: Good catch, just made that change!

PR #122, opened 2016-05-01
PEP8: Drastically reduce pep8 violations

jpanikulam:
Hunted them down


jpanikulam:
@mattlangford Pulling this first because it is already up to date

PR #126, opened 2016-05-07
BUILD: Fix CMakeLists issues, add cvx

jpanikulam:


simulation/sub8_gazebo/worlds/contact.world
```diff
@@ -84,14 +91,18 @@
           </geometry>
         </collision>
       </link>
-    </model>
+    </model> -->
+    <include>
```
mattlangford: I had only used the ground plane instead of the transdec to test my mission, did you want to have that be in master?

gnc/sub8_perception/nodes/marker_occ_grid.py
```diff
@@ -10,7 +10,6 @@
 from image_geometry import PinholeCameraModel
```
mattlangford, line 10: If you wanted to move this file as part of this PR, go ahead

simulation/sub8_gazebo/worlds/contact.world
jpanikulam: That's what I get for not being able to run Gazebo on my hometop

gnc/sub8_perception/nodes/marker_occ_grid.py
jpanikulam, line 10: ok, doing

simulation/sub8_gazebo/worlds/contact.world
jpanikulam: stupid github emails include the signature why

mattlangford: Side note, when I run the full simulator on my laptop, it runs at like .2 times normal speed which seems to mess up some component of the subs navigation.

jpanikulam: What happens? Jerky control, like we saw when we tried to do 5x speed simulation?

If so, try catkin_make-ing this latest pr and see if it helps

PR #132, opened 2016-05-14
Sub GNC without legacy Kill Master

zachgoins:
This PR is to remove any legacy kill handling from the current legacy GNC running on the sub. It also serves to start the mil_common folder which is going to be the uf_common of the new age. Anything shared between the sub and the boat will live in that folder. More will come in populating that folder in the future and moving some already implemented things to it but that will come in time.

**This commit does not add any kill handling to the system. At this time, the sub8.launch file starts the sub in the UNkilled state**

I am going to talk to Jake and see where the new alarm system should be woven in (preferably just one spot) and that will come shortly.


jpanikulam:
Note: Use git mv instead of cut/paste when moving files

jpanikulam:
Oops nevermind I guess this doesn't move, but is entirely seperate. In that case, let's come up with a way to avoid two c3's living in Sub. The one in Legacy has been nicely clang-formatted, takes like 30 secs to do (or you can just copy the fmting). 

Anyway, from here on, the "kill" scenario ought to reset c3. More importantly, if the kill alarm shuts down thrusters, RISE ought to listen to the Kill alarm as well, or we might get wacky behavior from the integrator term on unkill. 

To summarize conclusions from text conversation with zach earlier tonight:
- alarm-kill will go in thruster mapper at least
- Legacy kill will be expunged entirely (and the alarm-legacy-kill interface with it)
- Things common to boat and sub will live in sub for the foreseeable future.

Thoughts? Esp on the common?

Btw we should merge this after pool tomorrow

zachgoins:
Yeah I made this a new c3 since it does have some differences from the old one and I wanted to have a copy somewhere at least that was pure original c3. But maybe we can just let that live in software-common as a reference? I will get this one formatted.

Next PR will have kill alarm threaded into c3 and RISE. 

Also let me see why the build failed. 99% sure it is some messages that I forgot to set as a build first dependency in the CMake

zachgoins:
**Update:** I decided that it wasn't wise to pull this into the master without weaving kill handling in first. Will probably commit that today but still want to wait. Who knows, if something happens today and I can't add it and then you want to test tomorrow you'd be screwed.

zachgoins:
Okay everything I have addressed up to this point is taken care of. The new alarm system has been added in the proper places in Rise and the Thruster mapper and tested in the sim and everything has been clang formatted. 

**One thing:** I don't like that I had to do this but for some reason the AlarmListener class wasn't receiving messages and something wonky is going on so I accessed the "alarm_raise" topic directly. Which isn't good and I want to undo as fast as possible. The proper code is in there to access it through the class so I was thinking maybe someone else could take a look at it and see if maybe I am being blonde. I don't think so because I did it the exact same way on the boat and that boat's works fine but who knows. 

**To summarize:**
- mil_common c3 with new alarm system - clang formatted
- thruster mapper with new alarm system woven in 
- rise with new alarm system woven in

**Kill and Revive:**

```
rosrun sub8_alarm raise kill
rosrun sub8_alarm clear kill
```

zachgoins:
@jpanikulam @rleyva tear it up

jpanikulam:
Not building?

gnc/sub8_thruster_mapper/nodes/mapper.py
```diff
@@ -48,13 +49,24 @@ def __init__(self):
         self.B = self.generate_B(self.thruster_layout)
         self.min_thrusts, self.max_thrusts = self.get_ranges()
 
+        #self.kill_listener = AlarmListener('kill', self.kill_cb)
+        # Not the proper way to do this at all, but the kill listener class is being wonky right now
+        self.kill_listener = rospy.Subscriber('/alarm_raise', Alarm, self.kill_cb)
```
jpanikulam: Let's fix this so we don't have to hax -- could you describe more what went wrong? How were you publishing alarms that it didn't see?

When you were testing, did you run the alarm_handler node?

jpanikulam:
When duplicating existing stuff, like uf_common/msg, use git mv instead of copying. We should avoid having two copies of _anything_. Esp: most things won't behave properly when one is importing uf_common.msg.* and the action server is using sub8_msgs.msg.*

I am very excited to see this finally happen!

gnc/sub8_thruster_mapper/nodes/mapper.py
zachgoins: You are very right, the node wasn't running. My bad, I assumed it was in the master launch and never verified. But when I added it and started running it the alarm node gave me a bunch of errors so let me fix those and I will update this again. 

I will PR the alarm stuff separately. 

Getting closer still!

Also, weird reason for it not building it that is says "mil_c3_tgen" doesn't have a CMake. Looking into it.

jpanikulam: Uh-oh, what are the alarm errors? Looking at it, I see some rospy.logerr debugs that slipped through, but it runs dandily on my machine.

zachgoins:
So, everything is using the proper kill system now and I have added it to all launch files. There is one thing I just cannot figure out and I think it is a bug somewhere in the kill_handler.py file. When registering a new alarm using the raise and clear scripts you have to run it twice before the alarm starts taking effect. 

Ex. Start the sub, run clear kill, nothing happens but the alarm is registered. THEN the kill alarm works and calls the callback. But the first time it is registered it doesn't do anything. Ideally we would register and have it do an action at the same time. But finding that bug will take time.

I poured over the files but still haven't been able to find where it is happening. Not a huge deal but I want to find it none the less. 

For now just run clear kill once to register the alarm, then you are set to kill and unkill. 

@jpanikulam thoughts?

gnc/sub8_thruster_mapper/nodes/mapper.py
```diff
@@ -48,13 +49,23 @@ def __init__(self):
...
+        self.killed = True
+
         self.update_layout_server = rospy.Service('update_thruster_layout', UpdateThrusterLayout, self.update_layout)
         # Expose B matrix through a srv
         self.b_matrix_server = rospy.Service('b_matrix', BMatrix, self.get_b_matrix)
 
         self.wrench_sub = rospy.Subscriber('wrench', WrenchStamped, self.request_wrench_cb, queue_size=1)
         self.thruster_pub = rospy.Publisher('thrusters/thrust', Thrust, queue_size=1)
 
+    def kill_cb(self, msg):
```
jpanikulam, line 63: You don't need to check the alarm name

command/sub8_alarm/sub8_alarm/alarms/kill.py
```diff
@@ -1,23 +0,0 @@
-#!/usr/bin/env python
```
jpanikulam, line 1: Why is this file deleted?

legacy/rise_6dof/scripts/rise_6dof
```diff
@@ -15,8 +15,8 @@ from std_msgs.msg import Header
 from uf_common.msg import PoseTwist, PoseTwistStamped
 from sub8_ros_tools import pose_to_numpy, twist_to_numpy
 
-
-from kill_handling.listener import KillListener
+from sub8_alarm import AlarmListener
+from sub8_msgs.msg import Alarm
```
jpanikulam: unused import

legacy/rise_6dof/scripts/rise_6dof
```diff
@@ -15,8 +15,8 @@ from std_msgs.msg import Header
 from uf_common.msg import PoseTwist, PoseTwistStamped
 from sub8_ros_tools import pose_to_numpy, twist_to_numpy
 
-
-from kill_handling.listener import KillListener
```
jpanikulam, line 19: ITS HAPPENING!

jpanikulam:
Ok - code looks good except for minor comments.

For bookkeeping:

Repeated code **must** be removed from the repository. Please use git mv to move unchanged c3 files and uf_common messages instead of copying them. Your plan here is good, just nudging to follow convention.

jpanikulam:
ex: In one fel swoop we will pull this in and wipe out the legacy c3 package

This is especially a problem for messages, where the hash check will fail and we'll have conflicts all over the place.

zachgoins:
I think I addressed everything properly. I renamed everything form c3 that I didn't touch in git so it will keep tracking (my bad for not doing that from the start, just didn't know exactly what files I would have to change until I was totally done) and I removed imports and unneeded checks. 

About removing the kill file - Wasn't making a specific file for an alarm the old system before there was an AlarmListener class? The file isn't needed because we can listen for alarms anywhere so I figured it was best to not be confusing and remove it.

utils/sub8_ros_tools/CMakeLists.txt
```diff
@@ -9,3 +9,4 @@ if(CATKIN_ENABLE_TESTING)
     catkin_add_nosetests(test_ros_tools)
 endif()
 
+include_directories(sub8_cpp_includes)
```
jpanikulam, line 12: What is this?

utils/sub8_msgs/action/MoveTo.action
```diff
@@ -0,0 +1,12 @@
+# goal
```
jpanikulam: git mv, action requests are failing the hash check

jpanikulam:
Doesn't build

So Zach, the reason I am so adament about git mv'ing everything instead of cut/paste & copy/paste is stuff like [this line in RISE](https://github.com/uf-mil/Sub8/pull/132/files#diff-ba495a02d59a672e7a652a70cf73db93R15). Since c3 is now using a _different_ copy of the same message, missions crash as soon as a someone tries to move the sub.

I suggest running this with the full Gazebo simulation, you'll see a bunch of errors when you try to command the sub

Regarding removing the kill file: ok, gotcha

We're **alllllmooooost theeere**

mattlangford:
Any news here?

rleyva:
Any updates on this?

DSsoto:
Seems like this is dead for now, closing for good housekeeping. We can re-open when somebody claims this.

PR #134, opened 2016-05-15
BUILD: Add libompl-1.0.0 via MoveIt lib

rleyva:
Should fix @pemami4911's old OMPL version issue #67. MoveIt! uses OMPL 1.0.0. It'll suffice until we have our own repo.


rleyva:
Odd... Builds fine on my machine when the workspace is cleaned. I'll look into it.

jpanikulam:
Does it still work on your machine after removing libompl-dev and deleting
- catkin_ws/build
- catkin_ws/devel/shared
- catkin_ws/devel/include
- catkin_ws/devel/lib

?

rleyva:
I only removed the ompl from my install and ran catkin_make clean. I'll do a deeper purge and give it another go tomorrow.

PR #136, opened 2016-05-17
CVX: Un-Darsan the repository

jpanikulam:
Make the huge CVX autogenerated C files into single lines.


jpanikulam:
A while ago I committed tons of .c files that git was supposed to mark as binary, but it didn't. In response, I removed all of the human-readability whitespace to make them 1-6 lines each (From 26,000 lines before).

PR #137, opened 2016-05-18
Install Script Improvements and Fixes

sentree:
Things that were changed:
    - We no longer need to remove the stock ROS Gazebo packages to install Gazebo 7
    - There is now a default install location for no-argument running of the script
    - The DEPS_DIR was removed because it is not used by anything in the install process
    - OS verification ensures that everyone is on Ubuntu 14.04 (trusty)
    - Lines that added default trusty repositories are unnecessary and were removed
    - The script can now be run multiple times or for multiple workspaces without errors
    - The script no longer repeats one-time set up tasks that could cause errors

If the changes are accepted, the README will need to be updated to reflect the changes.
I could not find a function for the DEPS_DIR in the current build of the sub and it builds fine without it, but please let me know if I should add it back to the script.


install.sh
```diff
@@ -1,102 +1,136 @@
...
 rosdep update
 
-# Set up catkin directory
-mkdir -p "$CATKIN_DIR/src"
-cd "$CATKIN_DIR/src"
-catkin_init_workspace
-catkin_make -C "$CATKIN_DIR"
 
-# Set up sub
+#=====================================#
```
jpanikulam, line 91: Nice box

install.sh
```diff
@@ -1,102 +1,136 @@
...
+    mkdir -p "$CATKIN_DIR/src"
+    cd "$CATKIN_DIR/src"
+    catkin_init_workspace
+    catkin_make -C "$CATKIN_DIR"
+else
+    instlog "Using existing catkin workspace at $CATKIN_DIR"
+fi
+
+# If we're in the Semaphore-ci, we should run catkin_make in the actual build thread
+if  $SEMAPHORE; then
```
jpanikulam, line 107: It just occured to me that we could set an environment variable on Semaphore

rleyva:
Cool box

jpanikulam:
Great PR Anthony. Was pleased to see you used granular commits and followed the git guide.

sentree:
Thanks Jake, appreciate the feedback.

PR #141, opened 2016-05-18
SIM: Fix apparent regression of C3 behavior

jpanikulam:
- In the last 2 weeks, C3 stopped behaving properly with sim
- My theory:
  - We had some uncommitted local change on the Gazebo computer that made the sub more nimble

These changes make the sub less physically representative of reality (hence no changes to sub8_nocam), but c3 behaves as expected

### DIFF
--- a/simulation/sub8_gazebo/models/sub8/sub8.sdf
+++ b/simulation/sub8_gazebo/models/sub8/sub8.sdf
@@ -3,12 +3,13 @@
   <model name="sub8">
     <link name="base_link">
       <velocity_decay>
-        <linear>0.06</linear>
+        <linear>0.02</linear>
         <angular>0.06</angular>
       </velocity_decay>
 
       <inertial>
-        <mass>36.29</mass>
+        <!-- <mass>36.29</mass> -->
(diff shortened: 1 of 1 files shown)


mattlangford:
Before merging, I will try running with these params tonight and see how things go.

jpanikulam:
Ok. I had the same problem you were describing on my home machine, and this made it behave

mattlangford:
This seems to improve the issue but not entirely fix it.
I feel like when the trajectory projection (I guess the 'carrot on a stick') reaches the target, that's when the mission ends. Is that how it is supposed to behave?

jpanikulam:
That's how c3 is designed to behave, but that isn't really working for us. Making the sub more nimble in sim (droppping mass, reducing drag coeffs and damping) will make the sub more easily keep up with the carrot. Really, though, we need a better carrot.

mattlangford:
But the final position of the carrot is on the target, why does the sub get all confused when the mission ends? Does the mission ending kill some control element of the sub?

jpanikulam:
Closing for better solution

Issue #140, opened 2016-05-18
Make one-click simulated missions


jpanikulam:
- Take less than 5 mins to set up a gazebo test case
- Take less than 30 secs and 10 brain-cells (no cocaine) to start running an already existing test case

jpanikulam:
As in Semaphore, zero effort is expended to see if
- install script runs on a fresh install
- flake8 passes
- build works
- unit-tests and integration tests run

We should be able to do something similar in Gazebo. Being as painless as Semaphore is probably aiming too high, but minor improvements over @mattlangford 's current JobRunner implementation will save everybody days of test time.

jpanikulam:
What I think this would look like:
- CLI like missions

``` shell
# Internally runs/stops/restarts the appropriate mission as a subprocess
rosrun sub8_gazebo test test_gazebo
```

But I am just now trying to write a Gazebo job using @mattlangford 's script and finding it to be even easier than I thought

mattlangford:
So my idea was that the gazebo server would constantly be running the jub_runner, then people could start up missions anywhere on the network. That being said I'm sure it wouldn't be difficult to modify the script to be all in one and that would make it easier to use and would save dev time. The main thing my script would need is some identification over if the mission passed or failed (for bagging and statistics) which could be refactored by just returning information from the mission script.

sentree:
@mattlangford, what happened to the job_runner script? I am going to try to create a server on the gazebo machine to do 24/7 testing. Basically, there would be a web interface (or maybe just an SSH terminal interface if that takes too long) where jobs can be submitted, monitored, and controlled. A master queue would run the following in the specified order of priority:
- A specific set of tests required for every version of the master branch and pull requests (so we can compare the accuracy of each branch)
- Jobs submitted by individual users to be run a set number of times
- A series of tests that are to be cycled through when nothing else is being run

Basically, this will ensure there are no gaps in our software and provide us with cool statistics (@jpanikulam really wants a mileage ticker). Sound good? Any suggestions?

mattlangford:
It's in sub8_gazebo/diagnostics. I think a web interface would be cool.
We do have a mileage ticker in the gazebo controller.

mattlangford:
It might be easier to rosify the mission adding process and make an rqt plugin that lets you monitor mission status.

sentree:
[...]

PR #143, opened 2016-05-21
JOBRUNNER: Move to more mission-like API

jpanikulam:
- Jobs are called the same way missions are
- BagManager creates the bag folder if it doesn't already exist
- There is a bug with bag.dump that appears to have existed before
- C3 behaves a little better

Addresses #140


jpanikulam:
[And some documentation](https://github.com/uf-mil/Sub8/wiki/Gazebo-Job-Runner)

simulation/sub8_gazebo/diagnostics/job_runner.py
```diff
@@ -1,231 +1,118 @@
...
-        self.bagger.start_caching()
+        if self.record_log:
+            self.bagger.start_caching()
+
+        while current_loop < loop_count:
+            current_loop += 1
+            try:
+                yield current_job.setup()
+            except Exception, error:
+                print "On test #{}, a {} raised an error on test setup".format(
```
mattlangford: Should we log any errors like this or just print them out?

simulation/sub8_gazebo/launch/duck.launch
```diff
@@ -88,5 +88,10 @@
     </rosparam>
     <remap from="orig_odom" to="imu_odom"/>
   </node>
+
   <include file="$(find sub8_launch)/launch/subsystems/cameras/launch/stereo.launch" if="$(arg cameras)"/>
+
+  <group ns="/down/left">
```
mattlangford: Only start if cameras have been set to true?

simulation/sub8_gazebo/sub8_gazebo_tools/bag_manager.py
```diff
@@ -0,0 +1,148 @@
...
+        # Make bag directory if it doesn't exist
+        directory = os.path.join(self.diag_dir, 'bags')
+        if not os.path.exists(directory):
+            os.makedirs(directory)
+
+        bag = rosbag.Bag(os.path.join(directory, str(int(time.time())) + '.bag'), 'w')
+
+        gen = self.read_from_cache()
+
+        last_timestamps = dict(self.cache_dict)
```
mattlangford, line 102: Did adding this not fix it?

simulation/sub8_gazebo/diagnostics/job_runner.py
```diff
@@ -250,55 +136,61 @@ def run_job_loop(self):
...
-        '''
-        Reset sub position without breaking the simulator.
-        This could send a signal to the mission node so that it can set the sub position wherever it wants.
-        '''
-        print "Reseting Gazebo"
-        return
+    def queue_job(self, name, runs):
+        available_tests = [test_name for test_name in dir(gazebo_tests) if not test_name.startswith('_')]
+        assert name in available_tests, "Unknown test, {}".format(name)
+        assert isinstance(runs, int), "Cannot do non-integer runs, wtf are you trying to do?"
```
mattlangford, line 149: I JUST WANT TO RUN HALF THE MISSION

simulation/sub8_gazebo/sub8_gazebo_tools/bag_manager.py
```diff
@@ -0,0 +1,148 @@
...
+        self.cache_dict = {}
+        msgs = messages_to_bag['MESSAGES']
+        for msg in msgs.values():
+            # Get message information
+            msg_topic = msg['message_topic']
+            msg_type = msg['message_type']
+            msg_name = msg['message_name']
+
+            # Import the message
+            exec("from {0}.msg import {1}".format(msg_type, msg_name))
```
mattlangford, line 82: Did you want to change how the imports are done?

simulation/sub8_gazebo/sub8_gazebo_tools/bag_manager.py
```diff
@@ -0,0 +1,148 @@
...
+            for key in self.cache_dict:
+                msg = self.cache_dict[key].get_last_message()
+                msg_time = self.cache_dict[key].get_last_message_time()
+                if msg is not None:
+                    if msg_time < self.nh.get_time() - rospy.Duration(.5):
+                        # If the message is too old, forget about it.
+                        continue
+
+                    # Lang -- this doesn't work
+                    if msg_time == last_timestamps[key]:
```
mattlangford, line 115: What kind of error is it?

simulation/sub8_gazebo/sub8_gazebo_tools/bag_manager.py
jpanikulam, line 102: msgs still undefined

simulation/sub8_gazebo/sub8_gazebo_tools/bag_manager.py
jpanikulam, line 82: Nah, if you want to do it in the future you can use **import**, which is done in the **init**.py of gazebo_tests

simulation/sub8_gazebo/sub8_gazebo_tools/bag_manager.py
jpanikulam, line 115: A few lines later, it fails at msgs.append

I'm not sure if my guessery for what to do with last_timestamps is correct, either; is it?

simulation/sub8_gazebo/launch/duck.launch
jpanikulam: roger roger

simulation/sub8_gazebo/diagnostics/job_runner.py
jpanikulam: good point

simulation/sub8_gazebo/sub8_gazebo_tools/bag_manager.py
mattlangford, line 115: Yes I think that is right. It just needs to be a dictionary with the same keys as the caching dict so it can add timestamps to that.
It is supposed to dump to a bag here instead of save the messages (I copied this from the start_caching function but obv forgot to change some things).
Try this line instead of 119:
`bag.write(key, (yield msg), t=(yield msg_time))`

simulation/sub8_gazebo/diagnostics/gazebo_tests/align_marker_test.py
```diff
@@ -0,0 +1,85 @@
...
+from sub8_ros_tools import msg_helpers
+import missions
+
+from twisted.internet import defer
+import numpy as np
+
+from diagnostics.gazebo_tests import common
+
+
+class Job(common.Job):
```
mattlangford, line 16: This looks a hell of a lot cleaner then what I originally did

simulation/sub8_gazebo/diagnostics/job_runner.py
```diff
@@ -1,231 +1,118 @@
...
+
+        current_job = job_constructor(self.nh)
+
         current_loop = 0
 
-        self.bagger.start_caching()
+        if self.record_log:
+            self.bagger.start_caching()
+
+        while current_loop < loop_count:
```
mattlangford, line 68: Should initial setup be called up here somewhere?

simulation/sub8_gazebo/sub8_gazebo_tools/bag_manager.py
jpanikulam, line 115: Now getting a lot of "There's a problem recording /odom" printouts. /odom is definitely publishing, hopefully this isn't a weird tx problem

simulation/sub8_gazebo/diagnostics/job_runner.py
jpanikulam: We could print everything by default and let the user pipe the output into a file only if they desire?

mattlangford: Sure that works for me.
That would remove the lines below that dump the data to a file as well. Just make sure they are printed instead.

simulation/sub8_gazebo/sub8_gazebo_tools/bag_manager.py
mattlangford, line 115: That's happening on line 55 probably unless those printouts happen after there is a crash?
It should only print if the last message is none... hmmmm

simulation/sub8_gazebo/launch/duck.launch
```diff
@@ -2,7 +2,7 @@
   <param name="simulate" value="true" />
```
mattlangford, line 2: Hey I noticed I made a goof - this needs to be:
`<param name="/use_sim_time" value="true" />`
I think something used to set that but it has gotten removed.

jpanikulam, line 2: Don't worry - I think I wrote that line. /simulate is sub-specific and tells the thruster driver to behave differently. Gazebo automatically sets use_sim_time

simulation/sub8_gazebo/diagnostics/job_runner.py
jpanikulam, line 68: added

jpanikulam:
Addressed comments, couldn't figure out the /odom thing. In response, I added a --log switch, so you can turn off logging. Maybe you'll be able to solve it.

I decided against forcing the user to pipe output, because I think there are some things we want a log file for, and some things we want miscellaneous output for.

simulation/sub8_gazebo/launch/duck.launch
mattlangford, line 2: Hmm I was having issues with TF without it that were solved by adding the line.
Empty_world.launch would set it but I removed the call to that file. I think we need both.

mattlangford, line 2: Can the thruster mapper look for use_sm_time? Was there a reason for both?

jpanikulam, line 2: I am having those tf issues too, now. I think it's due to starting out of order.

One reason is that Forrest suggested using sim time even in real operation to force time synchronization

mattlangford, line 2: That's an interesting thought, I can't think of any reason not to as long as everything is using it

PR #147, opened 2016-05-25
MARKER: Update mission. Hydrophone driver too.

mattlangford:
Fixed marker mission based on info obtained at the pool.

Also pushing hydrophone driver for Jake to test.

**An interesting thing:**
Here is a graph of the rotation of the marker in the frame while using the non-energy optimized controller.
interesting
Very cyclic, more so then usual? Not sure.


gnc/sub8_perception/nodes/follow_orange_pipes.py
```diff
@@ -17,66 +17,162 @@ def __init__(self):
...
         self.image_sub = sub8_ros_tools.Image_Subscriber('/down/left/image_raw', self.image_cb)
         self.image_pub = sub8_ros_tools.Image_Publisher("down/left/target_info")
 
         self.occ_grid = MarkerOccGrid(self.image_sub, grid_res=.05, grid_width=500, grid_height=500,
                                       grid_starting_pose=Pose2D(x=250, y=250, theta=0))
 
-        self.range = sub8_ros_tools.get_parameter_range('/color/channel_guide')
-        self.channel_width = sub8_ros_tools.wait_for_param('/vision/channel_width')
+        #self.range = sub8_ros_tools.get_parameter_range('/color/channel_guide')
+        self.channel_width = 0.2  # sub8_ros_tools.wait_for_param('/vision/channel_width')
```
mattlangford, line 28: Should this stay hard coded? Does it ever change?

gnc/sub8_perception/nodes/follow_orange_pipes.py
```diff
@@ -17,66 +17,162 @@ def __init__(self):
...
+
+        center, angle_rad, [max_eigv, min_eigv] = self.get_pose(mask)
+
+        # Check if the box is too big or small.
+        xy_position, height = self.occ_grid.get_tf(timestamp=last_image_timestamp)
+        expected_area = self.occ_grid.calculate_marker_area(height)
+        if expected_area * .3 < rect_area < expected_area * 2:
+            cv2.drawContours(self.last_draw_image, [box], 0, (255, 255, 255), -1)
+        else:
+            angle_rad = 0
```
mattlangford, line 127: The idea here is that the marker is still in the frame just barley so we'll go to where we think it is, but the rotation is probably all wonky so just stay rotated like we are then fix once we get there.

gnc/sub8_perception/nodes/follow_orange_pipes.py
LucasBA, line 28: I don't think it changes. If it does i think it only changes from year to year.

jpanikulam, line 28: Necessary if god-president Trump mandates transition to imperial units

command/sub8_missions/missions/align_channel.py
```diff
@@ -10,49 +12,53 @@
...
 
+    # ---------------------------------------
+    # MOST OF THIS IS FOR THE SEARCH PATTERN.
+    # ---------------------------------------
     # This is just to get camera info
     resp = yield sub_singleton.channel_marker.get_2d()
     cam.fromCameraInfo(resp.camera_info)
 
-    intial_pose = sub_singleton.last_pose()
-    intial_height = yield sub_singleton.get_dvl_range()
```
jpanikulam, line 32: sidenote: we really out to make this better

command/sub8_missions/missions/align_channel.py
```diff
@@ -10,49 +12,53 @@
...
-    raidus = calculate_visual_radius(cam, intial_height)
```
jpanikulam, line 33: typo

command/sub8_missions/missions/align_channel.py
```diff
@@ -1,3 +1,5 @@
+from __future__ import division
```
jpanikulam, line 1: Jeez Matt, this mission is complicated!

command/sub8_missions/missions/align_channel.py
```diff
@@ -63,37 +69,38 @@ def go_to_marker(nh, sub_singleton, cam):
...
-            break
+        print "Target position: {}, Rotation: {}, Avg rotation: {}.".format(target_position, response.pose.theta, avg_rotation)
+        yield sub_singleton.move.set_position(target_position).yaw_left(avg_rotation).zero_roll_and_pitch().go()
 
-        last_target_position = target_position
-        yield nh.sleep(5.0)
+        yield nh.sleep(3.0)
 
 
 @util.cancellableInlineCallbacks
```
jpanikulam, line 106: We should make this a tx_sub function

Not for this pr

drivers/sub8_hydrophone_driver/nodes/hydrophone_driver.py
```diff
@@ -0,0 +1,122 @@
...
+    def request_data(self, srv):
+        '''
+        A polling packet consists of only a header and checksum (CRC-16):
+          HEADER     CHECKSUM
+        [  0x77  | 0x40 | 0x26 ]
+        '''
+        self.ser.flushInput()
+
+        message = HEADER
+        message += CRC(message)
```
jpanikulam: Does this run? Should it be self.CRC?

drivers/sub8_hydrophone_driver/nodes/hydrophone_driver.py
```diff
@@ -0,0 +1,122 @@
...
+
+        try:
+            self.ser.write(message)
+        except:  # Except only serial errors in the future.
+            self.disconnection_alarm.raise_alarm(
+                problem_description="Hydrophone board serial connection has been terminated."
+            )
+            return False
+
+        return self.listener
```
jpanikulam: I think you might be trying to return self.listener**()**?

drivers/sub8_hydrophone_driver/package.xml
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0"?>
+<package>
+  <name>sub8_hydrophone_driver</name>
+  <version>1.0.0</version>
+  <description>The sub8_hydrophone_driver package</description>
+
+  <maintainer email="Matthew.langford95@gmail.com">Matthew Langford</maintainer>
+
+  <license>MIL</license>
```
jpanikulam: Might mean MIT (MIL doesn't have a license, yet).

Not that important

command/sub8_missions/missions/align_channel.py
mattlangford, line 33: Could have an option to return the pose from the ground where it changed the z of the current pose. (meant to do this on the above comment)

drivers/sub8_hydrophone_driver/nodes/hydrophone_driver.py
mattlangford: Yep, should be

drivers/sub8_hydrophone_driver/nodes/hydrophone_driver.py
mattlangford: Yeah, I didn't actually run this last night since I didn't have the board.

drivers/sub8_hydrophone_driver/package.xml
mattlangford: We need a license!
How do we get one lol

command/sub8_missions/missions/align_channel.py
mattlangford, line 1: Without the search pattern I think it would be a lot more condensed. But oh well, bigger is better

PR #149, opened 2016-05-29
PERCEPTION: refactor and improve launch files

DSsoto:
Torpedo board pose pool adjustments, interfacing torpedo board mission code with perception code and cleaning up the sub8_perception package.


mattlangford:
@DSsoto do you see where the build failed?

DSsoto:
Yea, there were some artifacts of refactoring that I hadn't noticed.

DSsoto:
@jpanikulam or @mattlangford or anyone with admin acces to Semaphore, The Semaphore build script builds the sub repo twice, with the flag -DVISUALIZE being set to true and false. I deprecated that flag because nobody was using it. One of you should edit the build script so that the semaphore checks finish faster.

command/sub8_launch/launch/sub8.launch
```diff
@@ -8,6 +8,8 @@
     <arg name="depth" value="true" />
   </include>
 
+  <include file="$(find sub8_launch)/launch/subsystems/perception.launch" >
```
mattlangford: Do you need a '/' at the end of the line

command/sub8_missions/missions/torpedos.py
```diff
@@ -1,20 +1,35 @@
...
+@util.cancellableInlineCallbacks
+def align_to_board(msg, sub):
+    yield sub.move.set_position(target_position).zero_roll_and_pitch().go()
 
 @util.cancellableInlineCallbacks
 def run(sub):
 
     # dive to mission start depth
     print "descending to set course mission depth"
     mission_start_depth = 2.15        # meters
```
mattlangford: It may be a good idea to set this to a smaller value for now. If we accedently ran this mission in the pool we'd oops ourselves.

command/sub8_missions/missions/torpedos.py
```diff
@@ -1,20 +1,35 @@
...
-    sub.to_height(mission_start_depth)
+    sub.to_height(mission_start_depth, 0.5)
```
mattlangford, line 19: Should the height be negative? I would make a test mission and run it in the sim as to be sure of the sign.
Also it may be good for us to get some logic in all the height functions so that if a positive value gets passed to it, interpret it as a negative instead.

gnc/sub8_perception/nodes/torpedo_board_pose_est.py
```diff
@@ -22,105 +63,100 @@ def __init__(self):
...
         self.seq = req.pose_stamped.header.seq
-        self.stamp = req.pose_stamped.header.stamp
+        self.stamp = rospy.Time.from_sec(req.pose_stamped.header.stamp.to_time())
         self.frame_id = req.pose_stamped.header.frame_id
         self.position = req.pose_stamped.pose.position
         self.orientation = req.pose_stamped.pose.orientation
-        quat = np.array(
-            [self.orientation.x, self.orientation.y, self.orientation.z, self.orientation.w])
-        self.yaw = tf.transformations.euler_from_quaternion(quat, 'syxz')[0]
+        quat = np.array([self.orientation.x, self.orientation.y, self.orientation.z, self.orientation.w])
```
mattlangford, line 157: For future reference, sub8_ros_tools/msg_helpers has some functions to convert ros messages to numpy arrays.

gnc/sub8_perception/nodes/torpedo_board_pose_est.py
```diff
@@ -22,105 +63,100 @@ def __init__(self):
...
+        self.pose = np.array([self.position.x, self.position.y, self.position.z,
+                              tf.transformations.euler_from_quaternion(quat, 'syxz')[0]])
         self.left_cam_matx = np.array([
```
mattlangford, line 160: If you wanted to, you could change the service to include a camera_info message so that you wouldn't have to rebuild these matricies.

Also, was there a reason for moving the service to sub8_perception instead of back in sub8_msgs?

mattlangford:
I'll leave it to Jake to fix that, I'm not sure how

command/sub8_launch/launch/sub8.launch
DSsoto: fixed

command/sub8_missions/missions/torpedos.py
DSsoto: Changed it so it asks for input while we're pool testing

command/sub8_missions/missions/torpedos.py
DSsoto, line 19: I thought to_height took in a distance from the ground so I assumed positive values were what we wanted. I'll make sure to check.

gnc/sub8_perception/nodes/torpedo_board_pose_est.py
DSsoto, line 157: Cool, I'll keep it in mind for next time

DSsoto:
This represents the pose of the torpedo boar, not the camera. I'm also using an assumption of zero pitch and roll which is why this is required.

I moved all my services to sub8_msgs. Even got rid of the sub8_perception srv folder. Not sure why  git is not recognizing that change.

command/sub8_missions/missions/torpedos.py
mattlangford, line 19: Yeah that sounds correct, I always get confused about that.

command/sub8_missions/missions/torpedos.py
jpanikulam: Could do max(dvl_height + 1.5, 2.5)

mattlangford:
I would try to get the mission working in sim. If you need help, I will help next time I'm in town.

Did you still want me to work on the mission or did you want to finish it up?

command/sub8_missions/missions/torpedos.py
mattlangford: The mission runner should check if the target height would kill us or if it would make us go out of the water

DSsoto:
It's fine, I'll finish it up, thanks though.

The mission works fine in the sim up to this line: 
`yield sub.move.yaw_left(3.0).zero_roll_and_pitch().go()`
Sometimes it blocks forever, and sometimes it continues without moving the sub at all. I also tested it with different yaw values.

PR #150, opened 2016-06-03
Changes Made for the Clean Sub Install

sentree:
I made a few housekeeping changes to the install script. Most notably, we are now using environment variables to determine which platform we are running on instead of Jake's "spooky activity." I have set the relevant variables on the new sub OS and Semaphore.

I also made a script to manage hosts files. I will keep it up to date with the MIL router and it will automatically be run on the sub when the install script is run. If you run this, it should not interfere with any existing entries in the file or make duplicate entries. It will simply create a lock of entries at the end of the file and update it every time it is run.


sentree:
The Semaphore build tests assume that the sub will be installed under ~/repos/catkin_ws. Since I changed this to ~/sub_ws, I had to add -c ~/repos/catkin_ws to the ./install.sh command. This should not have any impact on other branches.

rleyva:
Note, if you ran the previous version of the install script you will have to give it the previous default ws name 'catkin_ws' with the -c parameter.

PR #151, opened 2016-06-04
SONAR: add pinger code location code, change interface to driver

DSsoto:
Driver now outputs 4D coordinates of pulse instead of raw timestamp data.


mattlangford:
Once this is working with the timestamps we got last night we can merge it in, I'd like to wait until then though.

DSsoto:
Got my code working with simulated timestamp data. I simulated pulses and generated timestamp data using our physical hydrophone configuration. My algorithm was able to very accurately estimate the position of the source. 

There is a little funkyness where the initial guess can  sometimes cause convergence to wrong values. It seems like the cost function function might not be globally convex. I will work on the math and try to iron this out, although it will not be a problem for most of the use cases for the sub.

However, I found out that Jake's board is producing wildly inconsistent timestamps that couldnt possibly represent accurate detection of the pinger's pulses by the four hydrophones. Locations that were close to the same were producing timestamps that were completely different from each other.

mattlangford:
Could you add the timestamps we collected at the pool with the relative positions of the pinger and hydrophones that we collected?

PR #152, opened 2016-06-05
ML: Additions to ML utilies.

mattlangford:
I might do some more with the segmenter to make it more friendly and fun!

PS: Sorry for any PEP8 errors I don't see, I don't have my sublime set up with all that yet. And don't know why it added the whole manual_grabcut.py file.


jpanikulam:
Nice. I think the manual_grabcut issue is because it got copied? Looks like there's two manual_grabcut.py's, one in labelling and one in Sub8_vision_tools

gnc/sub8_perception/sub8_vision_tools/machine_learning/verify_method_ada.py
```diff
@@ -0,0 +1,64 @@
...
+    }
+    for u_image, u_mask in data:
+        image = u_image[::2, ::2, :]
+        mask = u_mask[::2, ::2]
+        tic_observation = time()
+        some_observations = observe(image)
+        print '-------------------------'
+        print 'Observing took {} seconds'.format(time() - tic_observation)
+        tic_prediction = time()
+        segmentation = [x for x in [clf.predict(obs.reshape(1,-1)) for obs in some_observations]]
```
jpanikulam: Adaboost has numpy support, so if you reshape the some_observations to (-1, num_features) you can predict them all at once

Will be much faster

mattlangford:
Yep you're right, will fix today when I get home

mattlangford:
All the observations done at once is so much faster.

jpanikulam:
Super duper. Worthy of merge once build-check finishes.

I suggest trying thousands of convolutions using OpenCL. If you use local cacheing, you'll be able to get 100-1000x faster (as I'm sure you already know)

PR #158, opened 2016-06-11
ML: Add actual_wrench publisher

jpanikulam:
- Publish the actual actuated wrench (/wrench is unbounded and generally insane during normal operation)
- Add a mission for gathering motion data

**Image**: A sneak preview! Blue is what the sub actually did, green is what the estimated dynamics model says it should do, red is the error

dynamics_fitting

### DIFF
--- a/command/sub8_missions/missions/data_gather.py
+++ b/command/sub8_missions/missions/data_gather.py
@@ -1,0 +1,18 @@
+from twisted.internet import defer
+from txros import util, tf
+import numpy as np
+
+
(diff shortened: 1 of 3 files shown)


jpanikulam:
If you bag:

```
/wrench
/wrench_actual
/odom
/thrusters/thrust
/thrusters/thruster_status
/tf
```

While running data_gather or a mission like it, that would be sufficient @mattlangford

mattlangford:
I'm going to bed for pool day tomorrow, merge when checks pass.

PR #159, opened 2016-06-17
COMMAND: Add ability to run a mission autonomously

mattlangford:
Theres some various sim stuff and I fixed the annoying import errors we were getting too.

I can't reliably spell 'autonomously', so I used 'auto' everywhere instead

### DIFF
--- a/command/sub8_alarm/listeners/network_listener.py
+++ b/command/sub8_alarm/listeners/network_listener.py
@@ -4,12 +4,10 @@ from std_msgs.msg import String
 from std_srvs.srv import Empty, EmptyRequest
 from sub8_alarm import single_alarm
 from twisted.internet import reactor
-from missions import auto_mission
 
 
 class NetworkCheck(object):
-    def __init__(self, timeout=5.0, autonomous_msgs_req=100):
-        reactor.callWhenRunning(auto_mission.main)
+    def __init__(self, timeout=5.0, autonomous_msgs_req=20):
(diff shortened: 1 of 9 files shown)


command/sub8_alarm/listeners/network_listener.py
```diff
@@ -1,34 +1,59 @@
...
+        
+        # Make sure we don't accidentally let the sub loose.
+        # We need auto_msgs_req messages before we go autonomous mode.
+        self.auto_msgs_req = autonomous_msgs_req
+        self.auto_msg_count = 0
+
+        self.sub = rospy.Subscriber('/keep_alive', String, self.got_network_msg, queue_size=1)
+        self.auto_service = rospy.ServiceProxy('/go_auto', Empty)
         self.alarm_broadcaster, self.alarm = single_alarm('network-timeout', severity=1)
         rospy.Timer(rospy.Duration(0.01), self.check)
```
rleyva: Should we really be calling back 100 times a second?

rleyva:
At this moment it does not look like we will be using the gripper, so no need to add it for now.

command/sub8_alarm/listeners/network_listener.py
mattlangford: We have like 16 cores now, there are no rules. I'll dial it back in the morning

rleyva: ~0.25 sec be okay for this

PR #162, opened 2016-06-23
MULTIPLE: RVIZ, CAMERAS, GAZEBO changes

rleyva:
RVIZ: Default  to planar motion with spacenav mouse w/ option for full 6-dof if desired
RVIZ: Add distance from sub-to-waypoint when using shore control
RVIZ: Add depth from surface and depth from bottom measurements when using shore control
CAMERAS: Fixes Camera Info URL
CAMERAS: Add new downward camera launch (BlackFly) and deprecates old one (FireflyMV)
GAZEBO: Add barebones simulation in Gazebo (No visual, no cameras, no transdec items, etc) 
GAZEBO: Add navigation gate model

Notes: For the barebones launch make sure to run DISPLAY=:0 if you're SSH'ed into another machine

Also, sorry about the messy git history nonsense, I got feature branches messed up and had to do some serious mucking around to fix everything.


command/sub8_missions/tools/waypoint_utility.py
```diff
@@ -24,12 +27,32 @@ def __init__(self):
         When determining left and right, the wire faces away from you'''
         rospy.init_node('spacenav_positioning')
 
+        if len(sys.argv) == 0:
```
mattlangford: If you use a ros arg you can set a default in the launch file. If we want to add more driving modes later that might help, but it doesn't really matter now since you've already done it.

I think it should be a ros arg so you can set it when you launch shore control - do you think we will ever launch this file by itself?

rleyva:
Nope, you're right. This won't ever be launched solo. I'll fix an redo the PR

command/sub8_missions/tools/rviz_visualizer.py
```diff
@@ -35,7 +49,14 @@ def depth_callback(self, msg):
...
             id=0  # Keep these guys from overwriting eachother
         )
+        self.surface_marker.ns='sub'
+        self.surface_marker.header = sub8_utils.make_header(frame='/depth')
+        self.surface_marker.pose = marker.pose
+        self.surface_marker.text = str(round(distance, 3)) + 'm'
+        self.surface_marker.id = 0
+
         self.rviz_pub.publish(marker)
+        self.depth_marker.publish(self.depth_marker)
```
mattlangford, line 59: Should be `self.rviz_pub_t.publish(...)` ?

PR #163, opened 2016-06-25
SONAR: add testing program, change multilateration code

DSsoto:


forrestv:
Hey, FYI, if you're having trouble with solving the hydrophone localization
problem, I realized that the "Bancroft" algorithm, which is mentioned here
[1] as a way to solve the GPS multilateration problem also perfectly applies
to hydrophones.

It's a closed form algorithm that gives you 3D position, works for any
number of inputs >= 4, and is pretty simple.

I remember trying to implement an optimization-based solver for
hydrophones, and while I eventually got it to work reliably, it was a bit
unstable. Something like this should always work.

Tell me if you're interested. I could throw together a Python
implementation pretty quickly.

1: https://en.wikipedia.org/wiki/Global_Positioning_System#Solution_methods

DSsoto:
Our current plan was to add a fifth hydrophone so that the cost function became convex and could be solved for easily.

I've been able to reliably generate a 3D line that the source lies on with my current approach. So as a back-up plan,  I was going to repeat the process with the sub moving in between and calculate the "intersection" of those lines. However, I don't like that approach and would rather not have to go that route.

In other words, I'd definitely be interested!

forrestv:
Do you guys still have the old arrangement of the 4 (the T shape)?

DSsoto:
Sort of, with one of them out of the plane. The coordinates, in the hydrophone frame, are [here](https://github.com/uf-mil/Sub8/blob/master/command/sub8_launch/launch/subsystems/sonar.launch).

forrestv:
Is it possible to go back to the T or another planar arrangement? I'm just beginning to delve into it, but it seems like that arrangement (and potentially any other non-planar arrangement of 4) isn't ideal.

I'm pretty sure that:
- Any arrangement of 4 hydrophones results in two possible point solutions for at least some origin locations. With the arrangement you linked, seemingly exactly half of all origin locations result in ambiguous solutions.
- Arranging them all in a plane is nice because then the ambiguity is just a reflection along that plane, and we can always assume that the origin location is below us, resolving the ambiguity.

I think that moving back to a planar arrangement or moving to 5 hydrophones is required. The T arrangement before worked, but really a square is better. Also, the larger the better.

forrestv:
I do have the closed-form algorithm working fine, but I ran into this issue with the arrangement you gave me.

mattlangford:
I talked with Ken about possibly rearranging them into the T configuration, he says we can go back if we want.

drivers/sub8_sonar/nodes/plotter.py
```diff
@@ -0,0 +1,31 @@
...
+
+print "Waiting for sonar service"
+rospy.wait_for_service('~/sonar/get_pinger_pulse')
+sonar =  rospy.ServiceProxy('~/sonar/get_pinger_pulse', Sonar)
+print "sonar serv proxy created"
+#plt.xlim(-1.0, 1.0)
+#plt.ylim(-1.0, 1.0)
+
+try:
+	while(True):
```
mattlangford: Are these tabs or spaces?

If you're using sublime you can set `translate_tabs_to_spaces` to true in `Settings - Default`.

DSsoto:
The board design compatible with 5 hydrophones was recently finished. We're pushing to get the physical board operational before competition. I can also easily change the setup of the hydrophones, to a larger square. 

Forrest, could you explain how "Any arrangement of 4 hydrophones results in two possible point solutions for at least some origin locations"? 

Each of the hydrophone timestamps (difference in time of arrival measurement) constrains the source to half of a hyperboloid of two sheets with one of the foci at its respective hydrophone and the other at the reference hydrophone. Thus, the solution set with 4 hydrophones is the intersection set of 3 of those hyperboloids. I'm just not seeing how this necessitates the condition you described.

forrestv:
I did some testing with a square pattern; it will work fine. No real need to get 5 hydrophones running.

Sure, I can try to explain, but first I'll give an example so that you don't have to trust me:

```
>>> import numpy
>>> 
>>> r = 0.0254
>>> hydrophone_positions = map(numpy.array, [(0, 0, 0), (-r, 0, r), (r, 0, 0), (0, -r, 0)])
>>> 
>>> def get_relative_times_of_arrival(pos):
...     res = [numpy.linalg.norm(p - pos) for p in hydrophone_positions]
...     res = [r-res[0] for r in res]
...     return res
... 
>>> s1 = numpy.array([6.42057192, -14.94535233, -3.34558442])
>>> s2 = numpy.array([2.55726814,  -5.93892115, -1.30996347])
>>> print get_relative_times_of_arrival(s1)
[0.0, 0.014969530465872793, -0.0098037941328072975, -0.02285535964639962]
>>> print get_relative_times_of_arrival(s2)
[0.0, 0.014969530425247513, -0.0098037941123436667, -0.022855359659734731]
>>> assert numpy.all(numpy.isclose(get_relative_times_of_arrival(s1), get_relative_times_of_arrival(s2)))
>>>
```

forrestv:
Source is here: http://im.forre.st/pb/61675333.txt

Also, note that the problem we're solving is called "multilateration," not trilateration as I initially mentioned.

I find visualizing 4D hyperboloid surfaces hard, so I prefer to think of it as starting with four spheres, centered at the hydrophone locations and having radii of `speed_of_sound * arrival_time[i]`. Then, grow the spheres uniformly in radius until they all intersect at one point. However...

As the radii increase, two given spheres will start with no intersection, then intersect in a circle for the rest of time. Three given spheres will start with no intersection, then intersect in a pair of points for the rest of time (circle vs sphere = 2 points). As those two points follow their path, they have two chances to intersect with the fourth sphere, leading to two different solutions.

For some origin positions, one of those two solutions actually happens in the "past" - if you shrink the spheres instead of growing them. Another way to view it is that the two points already exist at t=0 and one is already inside the fourth sphere. And that kind of solution is possible to exclude, as it corresponds to the impossible situation of receiving the signal before it was broadcast. But, as testing and the above example showed, certain origin positions don't have this possibility.

If you want to look at this with hyperboloids, I think this problem arises because the hyperboloid sheets aren't at all like planes as you'd want to imagine. Instead, they're much more like pointy paraboloids, due to the close spacing of the foci. It's not hard to imagine that weird things happen when you're intersecting three of them. In 3D, I can imagine two intersecting in a shape that would be a closed loop and the third intersecting it at two points; I would think that would carry over to 4D.

forrestv:
Ah, I realize now that the hyperboloid sheets you mentioned aren't in 4D space at all, oops. What I said at the end still applies; just ignore the last sentence. I still think the spheres are easier to visualize.

mattlangford:
I'm going to merge this, I think the discussion can still continue afterwards.

DSsoto:
@forrestv I see what you mean, there's a lot of weird intersection curves that could occur when intersecting 3 hyperboloids. In the end, for competition it seems like we will be using a larger square arrangement and then after add the 5th hydrophone. We'll use Bancroft's algorithm too since it doesn't require 5 hydrophones to get the functionality we want. Could I get the implementation you put together so I don't waste time re-implementing it?

forrestv:
Sure, I'll clean it up and send it tonight.

DSsoto:
Cool, thanks.

forrestv:
Okay, here: https://gist.github.com/forrestv/96e33f1f63deb62afe79389f577b120f

Making it numerically stable took a bit of thinking and, ultimately, I didn't even find that clean of a solution, though it seems to work well. All the stuff having to do with `delta` is the hack that stabilizes it.

If you just run the script, it tests the algorithm against random inputs. It's been running for me for a while now without any failures.

The `estimate_pos` function is what you'll want to pull out; it should be obvious how to use it. Note that it returns a list of all the solutions it finds, so with a planar array it will return the true solution and the one reflected over the plane; you'll have to determine which one you want. With a planar array it will always return either two or zero solutions, but with a non-planar array it will sometimes return only one.

forrestv:
Also, I noticed something about the square configuration. It's not actually that great. A square configuration has a singularity along its axis - any pinger exactly along the axis perpendicular to the square and through its center will result in all of the times of arrival being the same, yielding no information about range. So, range will get increasingly inaccurate when the pinger is directly downward. That's not the end of the world; in previous years we completely ignored range information in the first place. The vector towards the pinger should still be good, I think...

The best configuration that I've found is three hydrophones uniformly distributed on a circle, then one at the center. Basically, a "Y". I've been running my script with that configuration and it looks great.

Sorry for flip-flopping on this.

forrestv:
Any problems understand/integrating that?

DSsoto:
I got it working perfectly in simulation now. I had to adjust it to work with non unit wave speeds. And it took me a while to figure out that I needed to express the hydrophone coordinates in smaller units to avoid large truncation errors that caused somewhat unstable results. You can see it  here:  #167.

I think for now, we might even be able to get away with our current hydrophone configuration even if your proposed "Y" could be somewhat better. It would be one less thing we need to do before competition. We'll do some testing this weekend and find out for sure.

Thanks for your help Forrest!

DSsoto:
@forrestv, when you were testing your Bancroft implementation did you ever try adding some noise to the simulated timestamps and seeing if it still converges to the right answer? 

For me, it works perfectly when I use simulated timestamps nearly all of the time, but if I add even a tenth of a microsecond of gaussian noise, it often fails to produce any result at all. I've been working on ways of getting around this, and was wondering if you had any suggestions.

image

image

DSsoto:
Btw, the above results were with this hydrophone configuration:

`<rosparam param="hydrophones">
    {   hydro0: {x:       0, y:       0, z: 0},
        hydro1: {x:       0, y:    25.4, z: 0},
        hydro2: {x:     -22, y:   -12.7, z: 0},
        hydro3: {x:      22, y:   -12.7, z: 0}}
</rosparam>`

Units are in millimeters and microseconds.

DSsoto:
@forrest, I found out that while the delta trick you added makes it so that we can find the source exactly pretty consistently given perfect timestamps, It makes it pretty much completely unusable for noisy timestamps. Removing it makes it so that we can get pretty good heading information about half the time and the other half we get no results.

I have an optimiztion based solver giving me nearly perfect heading nearly all of the time but I get no usable information about range.

forrestv:
If Bancroft algorithm doesn't give you any solutions, it means that there
_are no solutions_ and those timestamps are impossible (within the ideal
problem).

I wouldn't recommend removing the delta thing, at least not without
understanding it. It's there to make the B matrix full-rank, so that it can
be lstsq'd. If it isn't actually full rank, then I'm not sure exactly what
lstsq will do, but it'd be strange if it were what you wanted. I guess
testing could prove that it works, but there's probably a less hackish way.

We don't really need range anyway.

PR #165, opened 2016-06-29
SIM, RVIZ, COMMAND: Sonar sim and sub model to rviz

mattlangford:
The sonar sim is a WIP since the sonar algorithm is kind of up in the air right now. @DSsoto and I will have to confirm that it works.

**NOTE**
The rviz sub model shouldn't be used as a 100% accurate representation of the sub (since I made a low poly approximation of the sub) - so don't use it to precisely maneuver the sub around.


rleyva:
Everything looks good, I'll merge once it passes CI.

mattlangford:
Okay cool - might want to let @DSsoto look over the sonar simulator to see if everything there checks out.

rleyva:
Fair enough, he can merge it once he's satisfied. If it's not merged by tomorrow morning I'll merge it so that I can get the updated thruster config for testing.

DSsoto:
I dont have any merge, push privileges as of right now. @mattlangford  I'll look it over soon, I'm trying to integrate Forrest's bancroft implementation.

mattlangford:
Sweet, we can work on it over the weekend. Hopefully jake will be there too

gnc/sub8_thruster_mapper/launch/thruster_mapper.launch
```diff
@@ -24,7 +24,7 @@
             direction: [0, 0, -1]
           }
           FRL: {
-            node_id: 13,
+            node_id: 69,
```
DSsoto, line 27: Matt, thank you for your service!

simulation/sub8_gazebo/nodes/pinger_sim.py
```diff
@@ -44,7 +46,7 @@ def request_data(self, srv):
         pinger_pose = msg_helpers.pose_to_numpy(pinger_state.pose)[0]
 
         print sub_pose[1]
```
DSsoto, line 48: We shouldn't have print statements of raw data without labels

simulation/sub8_gazebo/nodes/pinger_sim.py
```diff
@@ -0,0 +1,71 @@
...
+
+class Pinger():
+    def __init__(self, wave_propagation_speed, precision=14):
+        self.get_model = rospy.ServiceProxy('/gazebo/get_model_state', GetModelState)
+
+        self.wave_propagation_speed = wave_propagation_speed
+        self.precision = precision
+
+        self.hydrophone_board_tf = np.array([0, 0, 0])
+        self.hydro_loc =  {'hydro0': {'x': 0,'y': 0,'z': 0},
```
DSsoto, line 21: You may want to use getParam for these, which are declared in the sonar launch file, they're likely to change.

simulation/sub8_gazebo/nodes/pinger_sim.py
```diff
@@ -0,0 +1,71 @@
...
+        hydro_poses = np.array([[self.hydro_loc['hydro0']['x'], self.hydro_loc['hydro0']['y'], self.hydro_loc['hydro0']['z']],
+                                [self.hydro_loc['hydro1']['x'], self.hydro_loc['hydro1']['y'], self.hydro_loc['hydro1']['z']],
+                                [self.hydro_loc['hydro2']['x'], self.hydro_loc['hydro2']['y'], self.hydro_loc['hydro2']['z']],
+                                [self.hydro_loc['hydro3']['x'], self.hydro_loc['hydro3']['y'], self.hydro_loc['hydro3']['z']]]) + hydro_board_pose
+        print hydro_poses
+        distances = np.linalg.norm(pinger_pose - hydro_poses, axis=1) 
+        timestamps = np.round(distances / self.wave_propagation_speed, decimals=self.precision)
+        print timestamps - timestamps[0]
+
+        # Don't forget estimated location is in map frame, transform to sub frame.
```
DSsoto, line 59: The coordinates in `hydro_poses` are based on the hydrophone board reference frame, therefore the estimate of the source will also be in that frame. Were you intending to have the coordinates of the pinger also represent a position w.r.t. the hydrophone frame?

We would need transform the results from hydrophone_board to sub .

As a side note, the convention is that  pose = position + orientation, usually doesn't refer to just points.

simulation/sub8_gazebo/nodes/pinger_sim.py
```diff
@@ -0,0 +1,71 @@
...
+        est_location = self.echo_locator.getPulseLocation(timestamps - timestamps[0])
+        est_location_np = rotate_vect_by_quat(np.array([est_location.x, est_location.y, est_location.z, 0]) - np.hstack((sub_pose[0], 0)), sub_pose[1])
+        est_location.x = est_location_np[0]
+        est_location.y = est_location_np[1]
+        est_location.z = est_location_np[2]
+
+        return est_location
+
+if __name__ == '__main__':
+    rospy.init_node('turbulator')
```
DSsoto, line 69: I assume you forget to change this?

simulation/sub8_gazebo/nodes/pinger_sim.py
```diff
@@ -0,0 +1,71 @@
...
+                           'hydro2': {'x': 0.0254, 'y': 0, 'z': 0},
+                           'hydro3': {'x': 0, 'y': -0.0254, 'z': 0} }
+
+        self.echo_locator = EchoLocator(self.hydro_loc, wave_propagation_speed)
+
+        # Wait for all the models and such to spawn.
+        rospy.Service('/sonar/get_pinger_pulse', Sonar, self.request_data)
+
+        rospy.sleep(3)
+        print self.request_data(None)
```
DSsoto, line 32: What's the reason for this?

simulation/sub8_gazebo/nodes/pinger_sim.py
mattlangford, line 48: Good point, I will try to not do this in the future

simulation/sub8_gazebo/nodes/pinger_sim.py
mattlangford, line 21: The sonar launch never gets launched for the sim right? That would mean we would have to change it in both launch files regardless.

I'm going to transition it over though.

simulation/sub8_gazebo/nodes/pinger_sim.py
mattlangford, line 32: When testing I would constantly restart the node - this made it easier to test quickly. I forgot to remove a bunch of the debugging stuff

simulation/sub8_gazebo/nodes/pinger_sim.py
mattlangford, line 59: The coordinates of the pinger should be in the hydrophone frame shouldn't they, isn't that how the data will be returned from the service?

simulation/sub8_gazebo/nodes/pinger_sim.py
DSsoto, line 21: We should just have a yaml file for the hydrophone locations that gets imported by any launch that needs it.

simulation/sub8_gazebo/nodes/pinger_sim.py
DSsoto, line 59: Thats right, just making sure. The important thing is that we are transforming the result from hydrophone to sub frame, not map to sub.

DSsoto, line 59: Actually, never mind. The pinger needs to be in sub frame because its an object that doesn't move with the sub. Then, every time we run the multilateration algorithm, we would first need to convert to hydrophone frame coordinates.

mattlangford, line 59: We can talk more in person and make sure everything is good

PR #167, opened 2016-07-01
SONAR, TORPEDOS, SIM

DSsoto:
The sonar software is consistently and accurately localizing simulated pulses from 4 differential timestamps. Credit goes to @forrestv  for his implementation of Bancroft's algorithm for multilateration. I expect that as soon as we can count on accurate timestamps from Jake's board, pinger localization will be a solved problem for us.
screenshot from 2016-07-01 09 54 53

Changed some of the torpedo code to make it more configurable.


drivers/sub8_sonar/sub8_sonar/sonar_driver.py
```diff
@@ -31,7 +34,7 @@ class Sub8Sonar():
     TODO: Add a function to try and reconnect to the serial port if we lose connection.
     TODO: Express pulse location in map frame
     '''
-    def __init__(self, hydrophone_locations, port, baud=9600):
+    def __init__(self, c, hydrophone_locations, port, baud=9600):
```
mattlangford: Should change default baud to 19200 (since that's what the board is talking at)

Also, this is something I should have done awhile ago but can you remove lines 95-99 and 101-106

drivers/sub8_sonar/sub8_sonar/sonar_driver_least_squares.py
```diff
@@ -0,0 +1,242 @@
+#!/usr/bin/env python
```
mattlangford: I think it would be good to refactor the two sonar drivers so that they both exist in the same file and the service call contains a string that allows the user to pick which method they want to use.

DSsoto: Done

drivers/sub8_sonar/launch/test.launch
```diff
@@ -1,10 +1,20 @@
 <launch>
```
mattlangford, line 1: Do we need this file?

drivers/sub8_driver_utils/camera_utils/src/dc1394_reset_bus.c
```diff
@@ -0,0 +1,67 @@
+//
```
mattlangford, line 1: Why are @rleyva's changes here?

drivers/sub8_sonar/sub8_sonar/sonar_driver.py
```diff
@@ -1,9 +1,13 @@
...
+from __future__ import division
+import math
+import numpy as np
+
 import rospy
 import rosparam
 
-import numpy as np
-from scipy import optimize
+import rospy
```
mattlangford, line 9: Multiple of the same imports

gnc/sub8_thruster_mapper/launch/thruster_mapper.launch
```diff
@@ -1,7 +1,7 @@
 <launch>
```
mattlangford, line 1: Is this the file Ralph wanted you push with this PR?

gnc/sub8_thruster_mapper/launch/thruster_mapper.launch.old
```diff
@@ -0,0 +1,110 @@
+<launch>
```
mattlangford, line 1: Do we need to keep old files like this - if the new thruster launch works

drivers/sub8_sonar/launch/test.launch
DSsoto, line 1: Yea, easy testing without messing with the actual sub hydrophone configuration.

drivers/sub8_driver_utils/camera_utils/src/dc1394_reset_bus.c
DSsoto, line 1: I think it's because he pushed some changes from his computer directly from the sub account, in between when I first opened this PR and now.

gnc/sub8_thruster_mapper/launch/thruster_mapper.launch.old
DSsoto, line 1: Ralph wants to keep it, just in case at least until competition.

drivers/sub8_sonar/sub8_sonar/sonar_driver.py
DSsoto, line 9: Fixed

gnc/sub8_thruster_mapper/launch/thruster_mapper.launch
DSsoto, line 1: No, he took care of it.

PR #172, opened 2016-07-19
OCR: Add tesseract ocr service

LucasBA:


gnc/sub8_perception/nodes/ocr_server.py
```diff
@@ -0,0 +1,85 @@
...
+        self.ocr_service = rospy.Service('vision/ocr', OcrRequest, self.request_characters)
+        self.image_sub = sub8_ros_tools.Image_Subscriber('/forward_camera/image_rect_color', self.image_cb)
+
+    def request_characters(self, srv):
+        self.white_list = srv.target_name
+        if (self.last_image != None):
+            print 'requesting', srv
+            response = self.ocr()
+
+            if response is False:
```
mattlangford: Should be 'if response is None', otherwise ocr() should return False if a letter isn't found, right?

gnc/sub8_perception/nodes/ocr_server.py
```diff
@@ -0,0 +1,85 @@
...
+
+    def __init__(self):
+        rospy.sleep(1.0)
+        self.last_image = None
+        self.last_draw_image = None
+        self.last_image_time = None
+        self.camera_model = None
+        self.white_list = None
+        self.ocr_service = rospy.Service('vision/ocr', OcrRequest, self.request_characters)
+        self.image_sub = sub8_ros_tools.Image_Subscriber('/forward_camera/image_rect_color', self.image_cb)
```
rleyva: This should be /stereo_front/(left or right)/image_rect_color

gnc/sub8_perception/nodes/ocr_server.py
LucasBA: Right, but actually that whole if else block is redundant. The None condition is fine. Ill fix that now.

mattlangford: Agreed

PR #173, opened 2016-07-26
PERCEPTION, SONAR, DRIVERS

DSsoto:
- Added current tested, and working multilateration code that can be adapted to work with new drivers.
- Added legacy sonar packages + adaptations to current sub configuration
- Added perception code to do  stereo object recognition and posing based on shape alone, assuming corrupted color information  

Includes some commented out dead code for easy debugging, and experimenting with different approaches while at competition.


drivers/hydrophones/scripts/hydrophones
```diff
@@ -0,0 +1,59 @@
+#!/usr/bin/env python
```
mattlangford, line 1: Is this all legacy code? If so in the future you may want to change the author on the commit to give credit to the original authors

DSsoto, line 1: I think credit for these legacy sonar pkgs goes to Matt Thompson.  How do I change the author?

mattlangford, line 1: Not sure if it's easy after you've committed ontop of it, but when you're commiting I think you can add '--author NAME'
I will review this PR on the plane to SD

PR #174, opened 2016-07-26
PERCEPTION: Move parameters to launch files.

mattlangford:
Also did some outlier detection


command/sub8_launch/launch/subsystems/perception.launch
```diff
@@ -1,18 +1,26 @@
...
   <arg name="color_calibration" default="default_color_calibration.yaml" />
 
   <rosparam file="$(find sub8_launch)/config/course_geometry.yaml" />
-  <!-- <rosparam file="$(find sub8_launch)/config/default_color_calibration.yaml" /> -->
   <rosparam file="$(find sub8_launch)/config/$(arg color_calibration)" />
 
-  <node pkg="sub8_perception" type="follow_orange_pipes.py" name="sub_orange_pipes" output="screen" />
-  <node pkg="sub8_perception" type="buoy_2d.py" name="buoy_2d" output="screen" />
+  <node pkg="sub8_perception" type="buoy_2d.py" name="buoys" output="screen">
+    <rosparam param="use_boost">True</rosparam>
```
DSsoto, line 9: It could be cleaner if we just create config files for our nodes and import them here so these launch files don't get so cluttered. I'm doing that for torpedo board stuff although I havent pushed it yet. Think about it for a future PR.

PR #175, opened 2016-07-26
COMMAND

DSsoto:
Improvements to launching and node configuration


command/sub8_launch/launch/sub8.launch
```diff
@@ -1,4 +1,5 @@
 <launch>
+  <rosparam param="/autonomous">False</rosparam>
```
mattlangford, line 2: Was there a reason to move this here instead of shore control?

command/sub8_alarm/nodes/network_keepalive.py
```diff
@@ -18,5 +18,5 @@ def keepalive_pub(self, *args):
 
 if __name__ == '__main__':
     rospy.init_node('network_keepalive')
-    ka = KeepAlive(rospy.get_param('~auto'))
+    ka = KeepAlive(rospy.get_param('/autonomous'))
```
mattlangford: In fact it might be better to do a check every time it publishes - that way we can go autonomous without restarting anything by just changing the param

command/sub8_launch/launch/sub8.launch
DSsoto, line 2: We can override the default false value from the command line before launching shore control.

command/sub8_alarm/nodes/network_keepalive.py
DSsoto: Good idea

DSsoto:
Closed because changes got integrated into #176

PR #186, opened 2017-01-08
INSTALLER: 16.04 and Kinetic integration.

RustyBamboo:
- Installs Catkin, Rosdep, binutils, Kinetic-desktop-full, OMPL, spacenav-node
- Downgrades Google's protobuf to compile Gazebo7's defaulted 2.6 protoc messages

Creating a pull request to encourage participation in developing and testing installer.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -101,7 +101,7 @@ if !($OS_CHECK); then
 
     # The script will not allow the user to install on an unsupported OS
     instwarn "Terminating installation due to incorrect OS (detected $DTETCTED_OS)"
-    instwarn "MIL projects require Ubuntu 14.04 (trusty)"
+    instwarn "This project requires Ubuntu Xenial (16.04)"
     exit 1
 fi
 
--- a/install.sh
+++ b/install.sh
@@ -62,7 +62,7 @@ done
 
 # The paths to the aliases configuration files
 BASHRC_FILE=~/.bashrc
-ALIASES_FILE=$CATKIN_DIR/src/.sub_aliases
+ALIASES_FILE=$CATKIN_DIR/src/sub8/.sub_aliases
 
 #==================#
 # Pre-Flight Check #
--- a/install.sh
+++ b/install.sh
@@ -62,7 +62,7 @@ done
 
 # The paths to the aliases configuration files
 BASHRC_FILE=~/.bashrc
-ALIASES_FILE=$CATKIN_DIR/src/sub8/.sub_aliases
+ALIASES_FILE=$CATKIN_DIR/src/Sub8/.sub_aliases
 
 #==================#
 # Pre-Flight Check #
@@ -210,10 +210,10 @@ catkin_make -C $CATKIN_DIR -B
 source ../devel/setup.bash
 
 # Clones the Sub8 directory unless it already exists
-if !(ls $CATKIN_DIR/src | grep --quiet "sub8"); then
+if !(ls $CATKIN_DIR/src | grep --quiet "Sub8"); then
     instlog "Downloading the Sub8 repository"
     cd $CATKIN_DIR/src
-    git clone https://github.com/uf-mil/sub8.git
+    git clone https://github.com/uf-mil/Sub8.git
     cd $CATKIN_DIR/src/Sub8
     git remote rename origin upstream
     instlog "Make sure you change your git origin to point to your own fork! (git remote add origin your_forks_url)"
@@ -268,6 +268,6 @@ cd $CATKIN_DIR/src
 git clone https://github.com/RustyBamboo/camera1394.git
 git clone https://github.com/txros/txros.git
 
-mv $CATKIN_DIR/src/sub8/gnc/sub8_trajectory_generator $CATKIN_DIR/src/sub8/gnc/.sub8_trajectory_generator
+mv $CATKIN_DIR/src/Sub8/gnc/sub8_trajectory_generator $CATKIN_DIR/src/Sub8/gnc/.sub8_trajectory_generator
 
 catkin_make -C $CATKIN_DIR -B
--- a/install.sh
+++ b/install.sh
@@ -203,7 +203,7 @@ else
     cd $CATKIN_DIR/src
 fi
 
-CMAKE_PREFIX_PATH=/opt/ros/kinetic
+CMAKE_PREFIX_PATH=$CMAKE_PREFIX_PATH:/opt/ros/kinetic
 git clone https://github.com/ros-drivers/driver_common
 catkin_make -C $CATKIN_DIR -B
 
--- a/install.sh
+++ b/install.sh
@@ -47,11 +47,22 @@ check_host() {
 }
 
 REQUIRED_OS="xenial"
+
+# Confirmation of where to put CATKIN directory
 CATKIN_DIR=~/mil_ws
+while true; do
+    instlog "The CATKIN Directory is $CATKIN_DIR. Is this okay(y/n)? "
+    read yn
+    case $yn in
+        [Yy]* ) break;;
+        [Nn]* ) read -p "Enter a new directory: " CATKIN_DIR; break;;
+        * ) echo "Please answer yes or no.";;
+    esac
+done
 
 # The paths to the aliases configuration files
 BASHRC_FILE=~/.bashrc
-ALIASES_FILE=~/.mil_aliases
+ALIASES_FILE=$CATKIN_DIR/src/.sub_aliases
 
 #==================#
 # Pre-Flight Check #
@@ -141,7 +152,7 @@ instlog "Installing Catkin and Rosdep"
 sudo apt-get update -qq
 sudo apt-get install -qq catkin python-rosdep
 
-instlog "Installing ROS Kinetic base packages"
+instlog "Installing the full ROS Kinetic"
 sudo apt-get install -qq ros-kinetic-desktop-full 
 
 # Source ROS configurations for bash on this user account
@@ -159,23 +170,27 @@ if !([ -f /etc/ros/rosdep/sources.list.d/20-default.list ]); then
 fi
 rosdep update
 
-instalog "Installing Sub8 ROS dependencies"
+instlog "Installing Sub8 ROS dependencies"
 sudo apt-get install -qq binutils-dev
 sudo apt-get install -qq ros-kinetic-ompl
 sudo apt-get install -qq ros-kinetic-spacenav-node
 
 # Get older version of Google protobuf for Gazebo
-instlog "Installing Google protobuf v2.6.1"
-wget https://github.com/google/protobuf/releases/download/v2.6.1/protobuf-2.6.1.tar.gz
-tar -xvzf protobuf-2.6.1.tar.gz
-cd protobuf-2.6.1
-./configure
-make
-make check
-sudo make install
-cd ..
-sudo rm -r protobuf-2.6.1
-sudo rm -r protobuf-2.6.1.tar.gz
+if !( protoc --version | grep "libprotoc 2.6.1" ); then
+    instlog "Installing Google protobuf v2.6.1"
+    wget https://github.com/google/protobuf/releases/download/v2.6.1/protobuf-2.6.1.tar.gz
+    tar -xvzf protobuf-2.6.1.tar.gz
+    cd protobuf-2.6.1
+    ./configure
+    make
+    make check
+    sudo make install
+    cd ..
+    sudo rm -r protobuf-2.6.1
+    sudo rm -r protobuf-2.6.1.tar.gz
+else
+    instlog "Detected correct version of protoc"
+fi
 
 # Set up catkin workspace directory
 if !([ -f $CATKIN_DIR/src/CMakeLists.txt ]); then
@@ -194,13 +209,21 @@ catkin_make -C $CATKIN_DIR -B
 
 source ../devel/setup.bash
 
-if !(ls $CATKIN_DIR/src | grep --quiet "Sub8"); then
+# Clones the Sub8 directory unless it already exists
+if !(ls $CATKIN_DIR/src | grep --quiet "sub8"); then
     instlog "Downloading the Sub8 repository"
     cd $CATKIN_DIR/src
     git clone https://github.com/uf-mil/sub8.git
     cd $CATKIN_DIR/src/Sub8
     git remote rename origin upstream
     instlog "Make sure you change your git origin to point to your own fork! (git remote add origin your_forks_url)"
+fi
+
+# Adds Sub8 aliases to $BASHRC_FILE unless already exists
+if !(cat $BASHRC_FILE | grep --quiet "source $ALIASES_FILE"); then
+    echo "" >> $BASHRC_FILE
+    echo "# Adds Sub8 aliases to shell environment" >> $BASHRC_FILE
+    echo "source $ALIASES_FILE" >> $BASHRC_FILE
 fi
 
 #================================#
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -1,81 +1,155 @@
...
+    NET_CHECK=false
+    echo -n "[ " && instfail && echo -n "] "
+fi
+echo "Internet connectivity check"
+
+if !($OS_CHECK); then
+
+    # The script will not allow the user to install on an unsupported OS
     instwarn "Terminating installation due to incorrect OS (detected $DTETCTED_OS)"
+    instwarn "MIL projects require Ubuntu 14.04 (trusty)"
```
DSsoto: This line is no longer applicable.

install.sh
```diff
@@ -1,81 +1,155 @@
...
 sudo apt-get update -qq
-if (env | grep --quiet "SUB=true"); then
-    sudo apt-get install -qq ros-indigo-desktop-full
-else
-    sudo apt-get install -qq ros-indigo-desktop
-fi
-sudo apt-get install -qq python-catkin-pkg python-rosdep gazebo7
+sudo apt-get install -qq catkin python-rosdep
+
+instlog "Installing ROS Kinetic base packages"
```
DSsoto: This is a desktop full install, not base packages. Let's not mislead people.

install.sh
```diff
@@ -85,55 +159,92 @@ if !([ -f /etc/ros/rosdep/sources.list.d/20-default.list ]); then
...
-
-#=====================================#
-# Workspace and Sub Repository Set Up #
-#=====================================#
+instalog "Installing Sub8 ROS dependencies"
+sudo apt-get install -qq binutils-dev
+sudo apt-get install -qq ros-kinetic-ompl
+sudo apt-get install -qq ros-kinetic-spacenav-node
+
+# Get older version of Google protobuf for Gazebo
```
DSsoto, line 178: Have you confirmed with others that this is necessary, or the best way to do it?

install.sh
```diff
@@ -85,55 +159,92 @@ if !([ -f /etc/ros/rosdep/sources.list.d/20-default.list ]); then
...
 else
     instlog "Using existing catkin workspace at $CATKIN_DIR"
+    cd $CATKIN_DIR/src
 fi
 
-# If we're in the Semaphore-ci, we should run catkin_make in the actual build thread
-if (env | grep --quiet "SEMAPHORE=true"); then
-    mv ~/Sub8 "$CATKIN_DIR/src"
-fi
+CMAKE_PREFIX_PATH=/opt/ros/kinetic
```
DSsoto: Could this potentially overwrite something important? Wouldn't it be better to just append?

install.sh
```diff
@@ -1,81 +1,155 @@
...
+
+            # Will return true if ping was successful and packet loss was below 25%
+            return `true`
+        fi
+    fi
+    return `false`
+}
+
+REQUIRED_OS="xenial"
+CATKIN_DIR=~/mil_ws
```
DSsoto, line 52: Should we prompt the user and have this as the default?

DSsoto (commented):
This is a good start. In my previous commit, I added a .sub_aliases file at the root of the repo, you should source this from the user's bashrc. I think we should also follow this configuration style for the other MIL repos.

We still need to find out how we're going to tie in CUDA into this. On machines not running already runing nvidia drivers this could be non-trivial.

install.sh
RustyBamboo: The gazebo messages are defaulted to protoc v2.6... The Ubuntu package is v3.0 (https://launchpad.net/ubuntu/+source/protobuf.) The different versions caused mine not to build. (Not sure if this is the case for others)

install.sh
```diff
@@ -62,7 +62,7 @@ done
 
 # The paths to the aliases configuration files
 BASHRC_FILE=~/.bashrc
-ALIASES_FILE=$CATKIN_DIR/src/.sub_aliases
+ALIASES_FILE=$CATKIN_DIR/src/sub8/.sub_aliases
```
DSsoto: The `s` in Sub8 should be capital.

install.sh
```diff
@@ -1,81 +1,166 @@
...
-CATKIN_DIR=~/sub_ws
-REQUIRED_OS="trusty"
+LOGPREFIX="${LOGCOLOR}INSTALLER:"
+WARNPREFIX="${WARNCOLOR}ERROR:"
+PASSTEXT="${PASSCOLOR}PASS"
+FAILTEXT="${WARNCOLOR}FAIL"
 
 instlog() {
-    printf "$GOODPREFIX $@ $NOCOLOR\n"
+    printf "$LOGPREFIX $@ $NOCOLOR\n"
```
sentree, line 14: I usually edit my bash scripts in nano, so naturally I prefer tabs to spaces. If there is a strong push to switch this to spaces, I recommend we use the bash standard of two spaces for indentation; however, I'd prefer to stick to tabs.

install.sh
```diff
@@ -1,81 +1,166 @@
...
+        # Uses packet loss percentage to determine if the connection is strong
+        if [ $HOST_PING -lt 25 ]; then
+
+            # Will return true if ping was successful and packet loss was below 25%
+            return `true`
+        fi
+    fi
+    return `false`
+}
+
```
sentree, line 48: Is there any particular reason you decided to do away with ros_git_get? Jake wrote the function to simplify the lines where git repositories were cloned and I was planning on using it's '--recursive' option to make our dependencies git submodules.

install.sh
```diff
@@ -1,81 +1,166 @@
...
+# Confirmation of where to put CATKIN directory
+CATKIN_DIR=~/mil_ws
+while true; do
+    instlog "The CATKIN Directory is $CATKIN_DIR. Is this okay(y/n)? "
+    read yn
+    case $yn in
+        [Yy]* ) break;;
+        [Nn]* ) read -p "Enter a new directory: " CATKIN_DIR; break;;
+        * ) echo "Please answer yes or no.";;
```
sentree, line 59: This could be confusing. Your code responds to values in [Y, y, N, n], but you request a 'yes' or 'no'.

install.sh
```diff
@@ -1,81 +1,166 @@
...
+    fi
+    return `false`
+}
+
+REQUIRED_OS="xenial"
+
+# Confirmation of where to put CATKIN directory
+CATKIN_DIR=~/mil_ws
+while true; do
+    instlog "The CATKIN Directory is $CATKIN_DIR. Is this okay(y/n)? "
```
sentree, line 54: Using a prompt based approach instead of script flags will probably work well for people that are new to MIL, but I will need to add a prompt for each flag in the common install script.

install.sh
```diff
@@ -1,81 +1,166 @@
...
+    case $yn in
+        [Yy]* ) break;;
+        [Nn]* ) read -p "Enter a new directory: " CATKIN_DIR; break;;
+        * ) echo "Please answer yes or no.";;
+    esac
+done
+
+# The paths to the aliases configuration files
+BASHRC_FILE=~/.bashrc
+ALIASES_FILE=$CATKIN_DIR/src/Sub8/.sub_aliases
```
sentree, line 65: It was a good idea to source the bash aliases file in the repository, but I will probably move some of these to software-common. I would also like to keep the file visible and in the scripts directory.

install.sh
```diff
@@ -1,81 +1,166 @@
...
-if (env | grep --quiet "SUB=true"); then
-    sudo apt-get install -qq ros-indigo-desktop-full
-else
-    sudo apt-get install -qq ros-indigo-desktop
-fi
-sudo apt-get install -qq python-catkin-pkg python-rosdep gazebo7
+sudo apt-get install -qq catkin python-rosdep
+
+instlog "Installing the full ROS Kinetic"
+sudo apt-get install -qq ros-kinetic-desktop-full
```
sentree, line 157: Good thinking here, we can finally do away with all of that metapackage nonsense!

install.sh
```diff
@@ -85,55 +170,104 @@ if !([ -f /etc/ros/rosdep/sources.list.d/20-default.list ]); then
 fi
 rosdep update
 
+instlog "Installing Sub8 ROS dependencies"
```
sentree, line 173: These should be under 'Dependency Installation'

install.sh
```diff
@@ -85,55 +170,104 @@ if !([ -f /etc/ros/rosdep/sources.list.d/20-default.list ]); then
...
     instlog "Using existing catkin workspace at $CATKIN_DIR"
+    cd $CATKIN_DIR/src
 fi
 
-# If we're in the Semaphore-ci, we should run catkin_make in the actual build thread
-if (env | grep --quiet "SEMAPHORE=true"); then
-    mv ~/Sub8 "$CATKIN_DIR/src"
-fi
+CMAKE_PREFIX_PATH=$CMAKE_PREFIX_PATH:/opt/ros/kinetic
+git clone https://github.com/ros-drivers/driver_common
```
sentree, line 207: What is this repository needed for? Also, I believe CMAKE_PREFIX_PATH is set when /opt/ros/kinetic/setup.bash is sourced.

install.sh
```diff
@@ -85,55 +170,104 @@ if !([ -f /etc/ros/rosdep/sources.list.d/20-default.list ]); then
...
+catkin_make -C $CATKIN_DIR -B
```
sentree, line 208: This build should only be done if the workspace has just been created, otherwise it will rebuild the user's existing workspace from scratch.

install.sh
sentree: Installing only ros-indigo-desktop-full on Semaphore saved a lot of time during builds, so I will have to add that back in.

install.sh
```diff
@@ -85,55 +170,104 @@ if !([ -f /etc/ros/rosdep/sources.list.d/20-default.list ]); then
...
+git clone https://github.com/txros/txros.git
 
-# Attempt to build the sub from scratch on client machines
-if !(env | grep --quiet "SEMAPHORE=true"); then
-    instlog "Building the sub's software stack with catkin_make"
-    catkin_make -C "$CATKIN_DIR" -j8
-fi
+mv $CATKIN_DIR/src/Sub8/gnc/sub8_trajectory_generator $CATKIN_DIR/src/Sub8/gnc/.sub8_trajectory_generator
+
+catkin_make -C $CATKIN_DIR -B
```
sentree, line 273: This line would result in building the repository twice on semaphore without a check.

install.sh
```diff
@@ -85,55 +170,104 @@ if !([ -f /etc/ros/rosdep/sources.list.d/20-default.list ]); then
...
+sudo apt-get install -qq sshfs
+sudo apt-get install -qq git-lfs gitk
+git lfs install --skip-smudge
+sudo apt-get install -qq tmux
+
+# Libraries needed by txros
+sudo apt-get install -qq python-twisted socat
+
+# Package management
+sudo pip install -q -U setuptools
```
sentree, line 248: You never install python-pip, which is not installed on Ubuntu server by default, so any pip installation would fail.

install.sh
```diff
@@ -85,55 +170,104 @@ if !([ -f /etc/ros/rosdep/sources.list.d/20-default.list ]); then
...
+    ./configure
+    make
+    make check
+    sudo make install
+    cd ..
+    sudo rm -r protobuf-2.6.1
+    sudo rm -r protobuf-2.6.1.tar.gz
+else
+    instlog "Detected correct version of protoc"
+fi
```
sentree, line 193: Are we sure there isn't a PPA for this?

install.sh
```diff
@@ -85,55 +170,104 @@ if !([ -f /etc/ros/rosdep/sources.list.d/20-default.list ]); then
...
+# Machine Learning
+sudo pip install -q -U scikit-learn > /dev/null 2>&1
+
+# Visualization
+sudo pip install -q -U mayavi > /dev/null 2>&1
+
+instlog "Cloning common GIT repositories"
 cd $CATKIN_DIR/src
-$CATKIN_DIR/src/Sub8/scripts/get_dependencies.sh
+git clone https://github.com/RustyBamboo/camera1394.git
```
sentree, line 268: Why is this necessary as opposed to installing ros-kinetic-camera1394?

install.sh
```diff
@@ -85,55 +170,104 @@ if !([ -f /etc/ros/rosdep/sources.list.d/20-default.list ]); then
...
 
-# Update the hosts file if necessary
-if (env | grep --quiet "SUB=true"); then
-    $CATKIN_DIR/src/Sub8/scripts/update-hosts.bash
+# Adds Sub8 aliases to $BASHRC_FILE unless already exists
+if !(cat $BASHRC_FILE | grep --quiet "source $ALIASES_FILE"); then
+    echo "" >> $BASHRC_FILE
+    echo "# Adds Sub8 aliases to shell environment" >> $BASHRC_FILE
+    echo "source $ALIASES_FILE" >> $BASHRC_FILE
 fi
```
sentree, line 228: I didn't see you move the Sub8 repository into the catkin workspace for Semaphore. If this step is not done, the master branch of Sub8 is what will be built, not the branch that was last updated.

sentree (changes requested):
Since we are moving toward a software-common repository, we are trying to build a unified MIL install script. I will be building and testing one on that repository over the next week and I will commit the changes from this PR that work well with you as the author. There are still a few questions I have before I add the changes, though.

install.sh
RustyBamboo: Standard is probably better

install.sh
RustyBamboo: I think it's a good thing to have

install.sh
RustyBamboo: I saw no real need to add more complexity. "git clone bleh" done. Although, it may be worth to add it back for the recursive option.

install.sh
RustyBamboo: Woops.

install.sh
RustyBamboo: The PPA is for the newer version (3.0.1 - i think.) The problem is that all the gazebo protoc messages don't specify that they're using 3.0.1, and so they default to 2.6.1.

install.sh
RustyBamboo: The driver_base in the drive_common is needed for the Camera1394 package.

Yeah, that seems to be the case...

install.sh
RustyBamboo: Yeah... the problem was that the Camera1394 depends on the driver_base... But during compilation Camera1394 would fail to build because it was looking for driver_base dependences which didn't exist yet. No clue how to fix this... My solution was to build driver_common first. Maybe build driver_common into the /ros/kinetic

install.sh
RustyBamboo: http://wiki.ros.org/camera1394... There doesn't seem to be support for kinetic...

install.sh
RustyBamboo: Ah... Smart.

install.sh
RustyBamboo: Semaphore is gud.

Not sure what the solution is for the camera1394 depending on driver_common without building twice...

install.sh
RustyBamboo: Although convenience is also good, which is probably better than following standards

install.sh
DSsoto: Make sure you keep any specific to Sub8 in that file.

sentree: That's the plan, a common aliases file and then one for each project installed.

DSsoto:
This was a great starting point but we have decided to move all of our install scripts to the installer repo.
Closing.

PR #189, opened 2017-01-11
ALARMS: Overhaul

mattlangford:
By using the new `ros_alarms` package, the alarms for the sub are now easier to use and understand.

More documentation can be found in the `ros_alarms` repository.

Everything in here is just removing old `sub8_alarms` and `kill_handling` packages and all the dependencies of them as well (in `c3_trajectory_generator`, `rise_6dof`, and various other little places where alarms are used)


kev-the-dev:
rviz visualizer is still broken in this. 26917a5eb03d4 has the fix so I can no longer bring great shame to family

DSsoto:
This built fine on my computer. Any ideas on why semaphore isn't building it?

Also, it seems to me like the sub8_trajectory_generator pkg has some pretty good code that can be built upon and it shouldn't be removed. If you want, we could move it to the legacy directory for the time being.

mattlangford:
Hmm, not sure. I'll look at that tomorrow. 

Okay, I will rebase those removes out and leave it there

mattlangford:
I think the install script is doing things wrong, @whispercoros can you see if the semaphore script does anything it shouldn't

DSsoto:
Yeah, the imu driver pkg didn't need to be built. And there's a wrong version of the gazebo ros pkgs that it tries to build.

I'm fixing that and we'll pull the install script from my fork on semaphore until @whispercoros has a stable version.

DSsoto (approved):
This is good stuff. It should build on semaphore now, will merge when it does.

DSsoto:
Actually, I forgot that sub8_trajectory_generator still depends on  sub8_alarms. If you make sub8_trajectory_generator a hidden folder, it will build and I will port it to ros_alarms in an upcoming PR.

PR #194, opened 2017-01-25
CAMERAS, LAUNCH, SIM, C3, PERCEPTION, POOL CHANGES

DSsoto:


mattlangford (dismissed):
Good luck today!

DSsoto:
Semaphore is saying that this isn't mergeable so it can't build it. Not really sure why because I'd expect github to complain if I did something wrong.

DSsoto:
This built on my computer, but not semaphore because I have some un-commited files. On my next PR to ros_alarms I will make sure this is fixed. Disregard this PR until then. 

Also added 2 small bugfix commits.

sentree:
I will get Semaphore working with the new install script today so that the Sub8 PRs can be merged.

simulation/sub8_gazebo/nodes/gazebo_controller.py
```diff
@@ -31,7 +31,7 @@ def __init__(self, target='sub8::base_link'):
 
         self.reset_srv = rospy.Service('gazebo/reset_gazebo', ResetGazebo, self.reset)
         self.state_pub = rospy.Publisher('odom', Odometry, queue_size=1)
-        self.world_state_pub = rospy.Publisher('world_odom', Odometry, queue_size=1)
+        self.map_state_pub = rospy.Publisher('map_odom', Odometry, queue_size=1)
```
mattlangford: I think this should be world. Map frame is zeroed at start, but the world frame is constant in gazebo (i.e. zero is the center of the world and not where the sub starts)

So it is the world frame not the map frame

A TF link between `map` and `world` that gets set on startup would be a nice addition.

sentree:
Have a look at [this](https://semaphoreci.com/uf-mil/sub8-2/branches/pull-request-194/builds/4). I was able to get this to build on Semaphore with a liberal amount of duct tape.

Full disclosure: I skipped building the simulation folder because I didn't want to deal with the level 10 wizardry necessary to get Gazebo 7 working with ROS Indigo. Since the changes to this directory in this PR are a python file and a world file, I don't think a build would have tested them anyway. As an important note, the failure point appears to be a ros_alarms test, so I'm not entirely sure what to make of that.

This fixes an issue that breaks the build in master and I've had to move all of the new software people over to David's fork just so they could get started. Everything builds on their machines as well.

mattlangford:
So what is the cause of this build error? It isn't a real thing (i.e. you can remove your build dir and build just fine)? 
I will merge after the comments are addressed if that is the case.

sentree:
Are you referring to the error on Semaphore or the error that users were getting with Sub8 master?

DSsoto:
@mattlangford I'm redoing some of the gazebo interface so that it is consistent with non-gazebo code (relies on map frame), but has both a notion of  world frame and map frame. I will add that to this PR if it hasn't been merged by then (no point in doing your suggested fix now since it would break non-sim code). I'm also going to transition our tf's to tf2 since it uses latched publishers for static transforms and it seems like you can have all the tf's in just one msg.

mattlangford:
Hmmm, not sure what you mean. Either take the gazebo changes out of here and put them all in another gazebo specific PR or like you said put the changes here.
No code running on the sub should need to worry about where it is in the world since everything should be map frame. Gazebo world frame shouldn't be mixed with a sub map frame except through a TF link

simulation/sub8_gazebo/nodes/gazebo_controller.py
DSsoto: Changed this back. Will be adding map frame stuff in an upcoming PR.

simulation/sub8_gazebo/diagnostics/gazebo_tests/common.py
```diff
@@ -15,7 +15,7 @@ class Job(object):
 
     def __init__(self, nh):
         self.nh = nh
-        self.true_pose_sub = nh.subscribe('/world_odom', Odometry)
+        self.true_pose_sub = nh.subscribe('/map_odom', Odometry)
```
mattlangford: Needs to be back to world odom for `set_model_position_test.py` to work

simulation/sub8_gazebo/diagnostics/set_model_position_test.py
```diff
@@ -57,10 +57,9 @@ def run_mission(srv, sub, nh):
 def main():
     nh = yield txros.NodeHandle.from_argv('test')
     world_position_actual = yield nh.subscribe('/world_odom', Odometry)
-    yield run_mission(None, world_position_actual, nh)
-    #nh.advertise_service('/gazebo/job_runner/mission_start', RunJob, lambda srv: run_mission(srv, world_position_actual, nh))
+    yield run_mission(None, map_position_actual, nh)
```
mattlangford: This variable doesn't exist

mattlangford (changes requested):

simulation/sub8_gazebo/diagnostics/set_model_position_test.py
DSsoto: @mattlangford Is there a reason that this file is not in `gazebo_tests/`? Also, this test and I imagine a lot of other stuff here is broken because it imports the old kill_handling package.

mattlangford: I think it was based on the old framework for setting up tests. All of the files in `gazebo_tests/` were set up with the new framework.
Trying to run anything from here would break due to the broken dependency with the old kill handling, you're right.

mattlangford (approved):

PR #197, opened 2017-01-29
Imaging Sonar

tessbianchi:
- SONAR: Adds package that allows the user to run and bag the sonar on the sub
- SONAR: changes CMAKE so that the sonar package doesn't build if the bvtsdk isn't found


gnc/sonar/include/BVWrapper.h
```diff
@@ -0,0 +1,48 @@
...
+
+#include <bvt_sdk.h>
+
+#include <opencv2/core/core.hpp> 
+#include <opencv2/highgui/highgui.hpp>
+#include <opencv2/imgproc/imgproc.hpp>
+#include <ros/package.h>
+
+namespace bv{
+	const int BW8 = 0;
```
mattlangford, line 12: Are these tabs?

sentree:
We seem to have gotten ourselves into a bit of a sticky situation with Semaphore using the old install script. I am looking in to setting up Docker on there to speed up our builds, but as a temporary fix, I will set it to use the installer repository once my PR is merged.

gnc/sonar/nodes/image_pub.cpp
```diff
@@ -0,0 +1,104 @@
...
+int main(int argc, char** argv){
+	try{
+		ros::init(argc, argv, "sonar_image_publisher");
+		ros::NodeHandle nh;
+		image_transport::ImageTransport it(nh);
+		image_transport::Publisher pub_bw = it.advertise("sonar/mono8", 1);
+		image_transport::Publisher pub_bw16 = it.advertise("sonar/mono16", 1);
+		image_transport::Publisher pub_color = it.advertise("sonar/color", 1);
+
+		const char host[] = "192.168.37.53";
```
mattlangford, line 22: You should pull this from a rosparam that defaults to this number

gnc/sonar/nodes/image_pub.cpp
```diff
@@ -0,0 +1,104 @@
...
+     	int old_slider = slider;
+     	int count = 0;
+
+		while (nh.ok()) {
+			std::clock_t begin = std::clock();
+			if(pings_not_received){
+				try{
+					sonar->reinitializeSonar();
+				}catch(char const* msg){
+					std::cout << "ERROR: " << msg << std::endl;
```
mattlangford, line 61: Probably should use ros logging errors here

gnc/sonar/prototype/p.py
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env python
```
mattlangford, line 1: Is this a test script? Does it need to be here

gnc/sonar/src/BVWrapper.cpp
```diff
@@ -0,0 +1,175 @@
...
+namespace bv{
+	bool BVWrapper::mapColors(BVTMagImage img, BVTColorImage& cimg){
+		if( this->mapper == NULL )
+		{
+			printf("BVTColorMapper_Create: failed\n");
+			return false;
+		}
+		
+		// Load the bone colormap
+		int ret = BVTColorMapper_Load(this->mapper, (ros::package::getPath("sonar") + "/res/jet.cmap").c_str());
```
mattlangford, line 12: The path could also be a rosparam.
I think this package should be called `sub8_imaging_sonar`, since `sonar` is ambiguous (since we currently have like two other sonar packages)

gnc/sonar/src/BVWrapper.cpp
```diff
@@ -0,0 +1,175 @@
...
+		BVTMagImage_GetHeight(img, &height);
+		int width;
+		BVTMagImage_GetWidth(img, &width);
+		const int arr_size = width * height;
+
+		BVTMagImage_GetBits(img, &this->bitBuffer);
+		unsigned short bitBufferCopy[arr_size];
+
+		// Copy the array over so that is now 'owned' by our program
+		std::copy(this->bitBuffer, this->bitBuffer + arr_size, bitBufferCopy);
```
mattlangford, line 40: Did you try using the `BVTMagImage_CopyBits` function?

gnc/sonar/src/BVWrapper.cpp
```diff
@@ -0,0 +1,175 @@
...
+		std::cout << "Sound Speed: " << speed << std::endl; 
+		float gain = -1.0;
+		BVTHead_GetGainAdjustment(this->head, &gain);
+		std::cout << "Gain: " << gain << std::endl;
+		int flags = -1;
+		BVTHead_GetImageFilterFlags (this->head, &flags);
+		std::cout << "Image Filter Flags: " << flags << std::endl;
+		int method;
+		BVTHead_GetImageProcessingMethod (this->head, &method);
+		std::cout << "Sound Speed: " << method << std::endl;
```
mattlangford, line 112: Should say `"Method: "` or something like that

mattlangford (changes requested):
Can you switch tabs to spaces, use rosparams where they are preferred to hardcoding (that would mean making a launch file as well), and rename the package so that it fits it better with the rest of the repo.
We'll have to wait to merge until the install script is sourced correctly to merge.

tessbianchi:
Great!

gnc/sonar/include/BVWrapper.h
sentree: Is the tab rebellion finally happening?

gnc/sonar/nodes/image_pub.cpp
sentree: I'd actually recommend doing a hostname lookup here. I don't know what the sonar's hostname is, but I will post it when I go into the lab later.

gnc/sonar/include/BVWrapper.h
DSsoto: No.

gnc/sonar/nodes/image_pub.cpp
mattlangford: Do you think the blueview sdk will accept a hostname? It might want the IP internally for some reason and not be happy with only the hostname. 
Or do you mean somehow look it up and pass that value to the sdk

gnc/sonar/src/BVWrapper.cpp
tessbianchi: Saw that after I wrote this.

gnc/sonar/include/BVWrapper.h
tessbianchi: tabs tabs tabs tabs tabs tabs
EVERYBODY!!

gnc/sonar/src/BVWrapper.cpp
tessbianchi: Yes

gnc/sonar/include/BVWrapper.h
mattlangford: https://google.github.io/styleguide/cppguide.html#Spaces_vs._Tabs
I'd be okay with 4 spaces, but 2 is the standard

mattlangford: wiki.ros.org/CppStyleGuide
"Indent each block by 2 spaces. Never insert literal tab characters."

DSsoto: Also, as per the [c++ ROS style guide](http://wiki.ros.org/CppStyleGuide/), braces should not be 'cuddled' (each brace should go in it's own line), and the contents of a namespace should not be indented.

gnc/sonar/include/BVWrapper.h
```diff
@@ -0,0 +1,48 @@
...
+	const int FILE = 4;
+	class BVWrapper{
+		private:
+
+			bool mapColors(BVTMagImage img, BVTColorImage& cimg);
+			bool getCVImages(BVTMagImage img, cv::Mat& image8, cv::Mat& image16, cv::Mat& image84);
+			void initializeSonar();
+
+			int dataSource = 0;
+			BVTHead head = NULL;
```
jpanikulam, line 25: One should use `nullptr` instead of `NULL`

It would be further wise to wrap each in a `std::shared_ptr<>`

DSsoto:
@tessbianchi any progress on this?

kev-the-dev:
Recommend closing as this driver was refactored and merged to mil_common

PR #200, opened 2017-03-08
REPO: Update references to uf-mil repositories after renaming

sentree:
This is just basic repository cleanup. References to software-common were updated to point to mil_common and a few important references to Sub8 were changed to point to SubjuGator. I initially considered changing all instances of "Sub8" to "SubjuGator", but that naming convention is so pervasive that it would take a significant time investment.


mattlangford (approved):

DSsoto:
@whispercoros Do you want to add the commits removing some of the stuff that is on its way out for **_The Great Repo Merge_** to this PR? And yea, don't try to change source code to use SubjuGator here, we'll do that slowly and carefully after this is done.

sentree:
When we do the merge, we will need to do some renaming and editing to make the files usable in a common context. We should only remove the files from the SubjuGator and NaviGator repositories once the files are fully merged and any files that depend on them are changed to work with the common versions.

PR #202, opened 2017-03-23
STARTGATE

RustyBamboo:
As requested, here it is.
Has worked on all the bags I've tried so far, however I can see many cases this will fail.

Overview:
- Finds mean and stddev of color. Then thresholds plus/minus deviations
- Contours, skeleton, and geometry checks
- Python mission is meh
- In Works: Better algorithm for perception...


command/sub8_missions/missions/start_gate.py
```diff
@@ -0,0 +1,116 @@
...
+    title="START_GATE", msg_color="white").fprint
+
+@txros.util.cancellableInlineCallbacks
+def run(sub_singleton):
+    fprint('Starting mission')
+    fprint('Assuming camera is facing gate')
+    fprint('Leveling off', msg_color="yellow")
+    yield sub_singleton.move.zero_roll_and_pitch().go()
+
+    start_gate_enable = yield sub_singleton.nh.get_service_client('/vision/start_gate/enable', SetBool)
```
mattlangford: You should set up the vision proxy, these 6 lines then become
```python
yield sub.vision_proxies.start_gate.start()
yield sub.vision_proxies.start_gate.get_pose()
```

command/sub8_missions/missions/start_gate.py
```diff
@@ -0,0 +1,116 @@
...
+def start_gate_find(sub_singleton):
+    global FOUND_START_GATE
+    start_gate_search = yield sub_singleton.nh.get_service_client('/vision/start_gate/pose', VisionRequest2D)
+    start_gate_search_res = yield start_gate_search(VisionRequest2DRequest(target_name=''))
+    FOUND_START_GATE = start_gate_search_res.found
+
+@txros.util.cancellableInlineCallbacks
+def search_pattern(sub_singleton):
+    if not FOUND_START_GATE:
+        fprint("Moving up .3", msg_color="yellow")
```
mattlangford: Won't we be at the surface when this mission is started?

command/sub8_missions/missions/start_gate.py
```diff
@@ -0,0 +1,116 @@
...
+    
+    fprint("Found start gate: " + str(start_gate_search_res.pose))
+    yield align_for_dummies(sub_singleton, start_gate_search_res)
+    fprint("YOLO")
+
+    while(FOUND_START_GATE):
+        start_gate_find(sub_singleton)
+        yield sub_singleton.move.forward(.3).zero_roll_and_pitch().go(speed=SPEED)
+
+    fprint("Finished")
```
mattlangford: Does it go through?

perception/sub8_perception/src/sub8_perception/start_gate.cc
```diff
@@ -0,0 +1,191 @@
+#include <sub8_perception/start_gate.hpp>
+Sub8StartGateDetector::Sub8StartGateDetector() : image_transport(nh), gate_line_buffer(30), running(true) {
+    image_sub = image_transport.subscribeCamera("stereo/right/image_raw", 1,
```
mattlangford: You should use the rectified image

perception/sub8_perception/src/sub8_perception/start_gate.cc
```diff
@@ -0,0 +1,191 @@
...
+
+void Sub8StartGateDetector::image_callback(const sensor_msgs::ImageConstPtr &image_msg,
+        const sensor_msgs::CameraInfoConstPtr &info_msg) {
+
+    if (running) {
+
+        cam_model.fromCameraInfo(info_msg);
+        image_time = image_msg->header.stamp;
+
+        if (gate_line_buffer.size() > 29) {
```
mattlangford: I think `gate_line_buffer.capactiy() - 1` would be better here, unless the 29 is a high precision machine learned number that optimally solves the problem.

perception/sub8_perception/src/sub8_perception/start_gate.cc
```diff
@@ -0,0 +1,191 @@
...
+
+    cv::findContours(canny_output, contours, hierarchy, CV_RETR_TREE, CV_CHAIN_APPROX_SIMPLE, cv::Point(0, 0));
+
+    // std::vector<cv::RotatedRect> minRect(contours.size());
+    // std::vector<cv::RotatedRect> minRect;
+    std::vector<cv::Rect> minRect;
+    for (size_t i = 0; i < contours.size(); i++) {
+        // cv::RotatedRect rect = cv::minAreaRect(cv::Mat(contours[i]));
+        cv::Rect rect = cv::boundingRect(cv::Mat(contours[i]));
+        if (rect.area() > 10000) {
```
mattlangford: Make arbitrary numbers params

mattlangford (approved):
Do it work?
I'll let David look at it as well

perception/sub8_perception/src/sub8_perception/start_gate.cc
RustyBamboo: Yes, I spoke with David and have ideas for smarter gate detection.

command/sub8_missions/missions/start_gate.py
RustyBamboo: Movement distances are arbitrary values... So I'm working on a way to get a 3D pose estimate and dynamically figure out distance

command/sub8_missions/missions/start_gate.py
RustyBamboo: Alright, will look into that

RustyBamboo:
The gate detection works fairly consistently on a lot of the bags, but I don't think it's robust enough... So I've been trying to find better ways to do this. I haven't had a chance to run the python mission, and will hopefully improve it with pose estimations.

sentree:
Do we have any good TRANSDEC bags of the start gate? If so, have you had successful runs with these? Also, seeing as the date of our next pool day remains an uncollapsed probabilistic wave function, I would recommend seeing if Ken can task someone on his team with modeling the gate for the simulator. That should give you a bit more insight than just testing on bags.

RustyBamboo:
I've been testing on a few bags from the TRANSDEC as well with fairly positive results. I'm trying to find bags that contain brighter colors (such as having buoys nearby) that might skew the distribution.

Good idea, I'll speak to Ken.

command/sub8_missions/missions/start_gate.py
```diff
@@ -0,0 +1,116 @@
...
+
+import numpy as np
+from twisted.internet import defer
+import os
+import yaml
+
+from sub8 import sub_singleton
+
+FOUND_START_GATE = False
+SPEED = 0.3
```
DSsoto: Consider making the code below part of a class and having the above as data members.

perception/sub8_perception/CMakeLists.txt
```diff
@@ -133,6 +133,10 @@ install(DIRECTORY include/${PROJECT_NAME}/
 #     "src/sub8_perception/*.cc"
 # )
 
+file(GLOB START_GATE_SOURCES
+      "src/sub8_perception/start_gate.cc"
```
DSsoto: C++ files should have extensions .cpp or .hpp, if you see any other files that violate this feel free to rename them.

perception/sub8_perception/CMakeLists.txt
```diff
@@ -133,6 +133,10 @@ install(DIRECTORY include/${PROJECT_NAME}/
 #     "src/sub8_perception/*.cc"
 # )
 
+file(GLOB START_GATE_SOURCES
```
DSsoto: You might not want to use GLOB, especially for just a couple of files. If a source file is added or removed but the CMakeLists is not changed, then it will not update and the build will fail.

GLOB is usually for gathering files that match an pattern, not for an explicit file. Try setting a variable called START_GATE_SOURCES instead.

perception/sub8_perception/CMakeLists.txt
```diff
@@ -199,6 +203,31 @@ target_link_libraries(
...
+add_dependencies(
+  sub8_start_gate
+   sub8_msgs_generate_messages_cpp
+   ${catkin_EXPORTED_TARGETS}
+)
+target_link_libraries(
+  sub8_start_gate
+    sub8_vision_lib
+    ${catkin_LIBRARIES}
+    ${PCL_COMMON_LIBRARIES}
```
DSsoto: You probably don't need to link to PCL libs, and we're not using bfd or dl anymore.

perception/sub8_perception/nodes/start_gate_node.cc
```diff
@@ -0,0 +1,23 @@
+#include <sub8_perception/start_gate.hpp>
+
+#include "ros/ros.h"
+
+#ifdef VISUALIZE
+  #warning(Compiling with visualization enabled)
```
DSsoto: Specify what node or this will not be useful.

perception/sub8_perception/src/sub8_perception/start_gate.cc
```diff
@@ -0,0 +1,191 @@
...
+        // rectanglePoints.push_back(std::vector<cv::Point2f>(rect_points, rect_points + sizeof rect_points / sizeof rect_points[0]));
+        rectanglePoints.push_back(std::vector<cv::Point2f>({minRect[i].tl(), minRect[i].br()}));
+    }
+
+    gate_line_buffer.push_back(rectanglePoints);
+}
+
+
+
+double Sub8StartGateDetector::distance() {
```
DSsoto: This could have a more descriptive name than distance.

perception/sub8_perception/src/sub8_perception/start_gate.cc
```diff
@@ -0,0 +1,191 @@
...
+    for (size_t i = 0; i < contours.size(); i++) {
+        // cv::RotatedRect rect = cv::minAreaRect(cv::Mat(contours[i]));
+        cv::Rect rect = cv::boundingRect(cv::Mat(contours[i]));
+        if (rect.area() > 10000) {
+            // std::cout << rect.area() << std::endl;
+            minRect.push_back(rect);
+        }
+    }
+    std::vector<std::vector<cv::Point2f>> rectanglePoints;
+    for (size_t i = 0; i < minRect.size(); i++) {
```
DSsoto: Use modern c++ man! 
```
for(auto& rect : minRect)
{
  rectanglePoints.push_back(std::vector<cv::Point2f>(rect.tl(), rect.br()));
}

```

perception/sub8_perception/src/sub8_perception/start_gate.cc
DSsoto: You should really look into using ROSCameraStream, it makes this trivial and handles a lot of the boilerplate for you and provides you with an image buffer in case you need temporal information. It also contains all the camera info stuff you need.

You can use it [like this](https://github.com/uf-mil/mil_common/blob/great_merge/perception/mil_vision/nodes/camera_stream_demo.cpp): `mil::ROSCameraStream<cv::Vec3b> left_cam_stream(nh, history_length);` @

perception/sub8_perception/src/sub8_perception/start_gate.cc
```diff
@@ -0,0 +1,191 @@
...
+    // cv::imshow("d", output);
+    cv::bitwise_not(output, output);
+    cv::dilate(output, output, cv::Mat(), cv::Point(-1, -1), 2, 1, 1);
+
+    cv::Mat skel(output.size(), CV_8UC1, cv::Scalar(0));
+    cv::Mat temp(output.size(), CV_8UC1);
+
+    cv::Mat element = cv::getStructuringElement(cv::MORPH_CROSS, cv::Size(3, 3));
+
+    bool done;
```
DSsoto: `done` has an undefined value, as does the variable `max` below. You should always initialize your variables when you declare them unless you're sure that you won't be generating some undefined behavior.

perception/sub8_perception/src/sub8_perception/start_gate.cc
```diff
@@ -0,0 +1,191 @@
...
+            cv::imshow("ay", show);
+            cv::waitKey(30);
+        }
+        findGate(image_msg);
+
+    }
+}
+
+
+void Sub8StartGateDetector::findGate(const sensor_msgs::ImageConstPtr &image_msg) {
```
DSsoto: You should comment this so that someone could be able to get the gist of what your algorithm is trying to do when glancing at the code.

perception/sub8_perception/src/sub8_perception/start_gate.cc
```diff
@@ -0,0 +1,191 @@
...
+        const sensor_msgs::CameraInfoConstPtr &info_msg) {
+
+    if (running) {
+
+        cam_model.fromCameraInfo(info_msg);
+        image_time = image_msg->header.stamp;
+
+        if (gate_line_buffer.size() > 29) {
+
+            for (size_t i = 0; i < gate_line_buffer.size(); i++) {
```
DSsoto: Since you're not doing arithmetic with the indexing variable, you should use range-based for loops.

DSsoto (changes requested):
Requesting some changes. Main points:
+ Make your code conform to the [ROS style guide](http://wiki.ros.org/CppStyleGuide) (2 spaces, no cuddle braces, underscores after class data members, etc.)
+ You should try to use more descriptive variable naming and add descriptive comments. Try to make the code easier to read for others.
+ Try to use more features of modern c++, they will make your code safer and more concise.
+ Avoid using magic constants

I'd like to see how your code runs on bags sometime, and I can help you get the full pose of the gate if the you were able to get the silhouette quadrilateral successfully.

perception/sub8_perception/CMakeLists.txt
kev-the-dev: *pushes back glasses* umm actually the [official ros style guide](http://wiki.ros.org/CppStyleGuide#Files) states that c++ header files should have the extension **.h**

DSsoto: Haha, let's not follow that part of the guide because it is stupid.

sentree: *Removes glasses dramatically* I kind of have to agree with @ironmig here, the style guide is clear and there's no good reason to ignore this part of it.

kingkevlar: _Breaks glasses_

samfisher47: *Gouges eyes out*

DSsoto: _Throws to the ground and kicks_ Me and @ironmig at the very least agreed that the good reason to ignore this part of it is because .h is for c files while .hpp is for c++.

mattlangford: Let's actually move to no header files, they're for the weak.

DSsoto: Re-write this in assembly, compiled languages are for the weak.

sentree: If we all wrote in assembly, there would be no compilation errors and we would get better performance. I'm for this.

command/sub8_missions/missions/start_gate.py
```diff
@@ -21,96 +21,98 @@
...
-        fprint("Moving right .3", msg_color="yellow")
-        yield sub_singleton.move.right(.3).zero_roll_and_pitch().go(speed=SPEED)
+    @txros.util.cancellableInlineCallbacks
+    def run_mission(self):
+        fprint('Starting mission')
+        fprint('Assuming camera is facing gate')
+        fprint('Leveling off', msg_color="yellow")
+        yield self.sub_singleton.move.zero_roll_and_pitch().go()
+        fprint('Going down', msg_color="yellow")
+        yield self.sub_singleton.move.down(1).go()
```
DSsoto: This seems arbitrary since you don't know at what depth you're starting, what's the reason?

command/sub8_missions/missions/start_gate.py
```diff
@@ -21,96 +21,98 @@
...
-    fprint("YOLO")
-
-    while(FOUND_START_GATE):
-        start_gate_find(sub_singleton)
-        yield sub_singleton.move.forward(.3).zero_roll_and_pitch().go(speed=SPEED)
-
-    fprint("Finished")
+class StartGateMission:
 
+    FOUND_START_GATE = False
```
DSsoto: `FOUND_START_GATE` isn't actually a constant.

Can you make speed a ros param?

command/sub8_missions/missions/start_gate.py
```diff
@@ -21,96 +21,98 @@
...
-    fprint("Found start gate: " + str(start_gate_search_res.pose))
-    yield align_for_dummies(sub_singleton, start_gate_search_res)
-    fprint("YOLO")
-
-    while(FOUND_START_GATE):
-        start_gate_find(sub_singleton)
-        yield sub_singleton.move.forward(.3).zero_roll_and_pitch().go(speed=SPEED)
-
-    fprint("Finished")
+class StartGateMission:
```
DSsoto: > A "New Class" is the recommended way to create a class in modern Python.

`class StartGateMission(object):`

command/sub8_missions/missions/start_gate.py
```diff
@@ -21,96 +21,98 @@
...
+
+        start_gate_search = yield self.sub_singleton.nh.get_service_client('/vision/start_gate/pose', VisionRequest2D)
+        fprint("Searching for start gate pose")
+        start_gate_search_res = yield start_gate_search(VisionRequest2DRequest(target_name=''))
+
+        if not start_gate_search_res.found:
+            fprint("Waiting a few seconds and trying again:")
+            time.sleep(3)
+            start_gate_search_res = yield start_gate_search(VisionRequest2DRequest(target_name=''))
+        # yield start_gate_enable(SetBoolRequest(data=False))
```
DSsoto: Dead code

DSsoto (commented):
I'll be ready to merge after these minor comments are addressed.

Could you add a unit test to this for your next PR?

perception/sub8_perception/src/sub8_perception/start_gate.cpp
```diff
@@ -0,0 +1,204 @@
+#include <sub8_perception/start_gate.hpp>
+Sub8StartGateDetector::Sub8StartGateDetector() : image_transport_(nh_), gate_line_buffer_(30), running_(true)
+{
+  image_sub_ =
+      image_transport_.subscribeCamera("stereo/left/image_color", 1, &Sub8StartGateDetector::imageCallback, this);
```
DSsoto: This needs to be image_**rect**_color

perception/sub8_perception/src/sub8_perception/start_gate.cpp
```diff
@@ -10,66 +10,60 @@ Sub8StartGateDetector::~Sub8StartGateDetector() {}
...
                     continue;
                 }
+
+                //If the size of the element is greater than one deviation from the mean, then delete
                 if (abs(s.x - mean(accSizeX)) > sqrt(variance(accSizeX)) ||
                         abs(s.y - mean(accSizeY)) > sqrt(variance(accSizeY))) {
-                    gate_line_buffer[29].erase(gate_line_buffer[29].begin() + i);
+                    gate_line_buffer[gate_line_buffer.capacity()-1].erase(gate_line_buffer[gate_line_buffer.capacity()-1].begin() + i);
                     i--;
                     std::cout << "Deleted size: " << s << std::endl;
```
DSsoto: `ROS_INFO_STREAM("Deleted size: " << s)` You shouldn't just use cout from code running on the robot.

perception/sub8_perception/src/sub8_perception/start_gate.cpp
```diff
@@ -10,66 +10,60 @@ Sub8StartGateDetector::~Sub8StartGateDetector() {}
...
-                // }
-                cv::rectangle(show, gate_line_buffer[29][i][0], gate_line_buffer[29][i][1], cv::Scalar(255, 0, 0), 2, 8, 0);
+            //Visualize on of the saved gates in the buffer
+            
+// #ifdef VISUALIZE
+            cv::Mat show = cv_bridge::toCvShare(image_msg, "bgr8")->image;
+            for(auto &gate : gate_line_buffer[29]) {
+                cv::rectangle(show, gate[0], gate[1], cv::Scalar(255, 0, 0), 2, 8, 0);
             }
             cv::imshow("ay", show);
```
DSsoto: Instead of using #ifdef visualize and opening gui windows, it would be cleaner to publish visualization images.

perception/sub8_perception/src/sub8_perception/start_gate.cpp
```diff
@@ -0,0 +1,204 @@
...
+
+        // If the size of the element is greater than one deviation from the
+        // mean, then delete
+        if (abs(s.x - mean(accSizeX_)) > sqrt(variance(accSizeX_)) ||
+            abs(s.y - mean(accSizeY_)) > sqrt(variance(accSizeY_)))
+        {
+          gate_line_buffer_[gate_line_buffer_.capacity() - 1].erase(
+              gate_line_buffer_[gate_line_buffer_.capacity() - 1].begin() + i);
+          i--;
+          std::cout << "Deleted size: " << s << std::endl;
```
DSsoto: This should be `ROS_INFO_STREAM("Deleted size: " << s)`

Don't use cout from ros nodes, so that we can control the verbosity of the output.

perception/sub8_perception/src/sub8_perception/start_gate.cpp
```diff
@@ -0,0 +1,204 @@
...
+
+      // Visualize on of the saved gates in the buffer
+
+      // #ifdef VISUALIZE
+      cv::Mat show = cv_bridge::toCvShare(image_msg, "bgr8")->image;
+      for (auto &gate : gate_line_buffer_[29])
+      {
+        cv::rectangle(show, gate[0], gate[1], cv::Scalar(255, 0, 0), 2, 8, 0);
+      }
+      cv::imshow("ay", show);
```
DSsoto: Instead of using defines for visualization, the cleaner way would be to publish debug images.

DSsoto:
@RustyBamboo any updates on this? It would be cool to merge before pool day tomorrow.

RustyBamboo:
I would like to see if the sonar makes a usable pointcloud for the gate and implement that into this mission.
I'll work on a unittest.

DSsoto (approved):

PR #204, opened 2017-04-03
PATH MARKER, CAMERAS, TF

kev-the-dev:
- Rewrite of path marker perception, including 3d pose estimation using solvepnp and unittests
- some fixes in mission system discovered during testing/sim
- after discussion with @DSsoto, change frame ids/namespaces of cameras to increase sanity. This may cause some issues with existing software, but is a good change to make
-- Cameras are under /camera (/camera/front/left, /camera/front/right, /camera/down/left)
-- frame ids for camera match those in tf, now front_left_cam, front_right_cam, and down_left_cam


kev-the-dev:
btw, we need a solution for unit test data. @whispercoros, what you think? [Catkin can download it for you](http://docs.ros.org/kinetic/api/catkin/html/howto/format2/downloading_test_data.html) with an http link, we could write our own scripts, git lfs, etc.

command/sub8_launch/launch/tf.launch
```diff
@@ -1,19 +1,19 @@
 <launch>
   <!-- This is not exactly right, because of weirdness in the K-matrix from camera_info for the stereo cams -->
     <node pkg="tf" type="static_transform_publisher" name="stereo_front_broadcaster"
```
mattlangford, line 3: Any research into TF2? It does these static broadcasters way better with much less issues on playback

perception/sub8_perception/nodes/follow_orange_pipes.py
```diff
@@ -4,315 +4,228 @@
...
+from sub8_ros_tools import Image_Publisher, Image_Subscriber
 from image_geometry import PinholeCameraModel
-
-SEARCH_DEPTH = .65  # m
+from visualization_msgs.msg import Marker
 
 rospack = rospkg.RosPack()
 
-class MarkerFinder():
+'''
```
mattlangford: Good overview comments, but you need some in the code as well

perception/sub8_perception/nodes/follow_orange_pipes.py
```diff
@@ -4,315 +4,228 @@
...
+            cv2.line(self.last_image, (refs[0][0][0], refs[0][0][1]), (refs[1][0][0], refs[1][0][1]), (0, 0, 255)) # X axis refs
+            cv2.line(self.last_image, (refs[0][0][0], refs[0][0][1]), (refs[2][0][0], refs[2][0][1]), (0, 255, 0)) # Y axis ref
+            cv2.line(self.last_image, (refs[0][0][0], refs[0][0][1]), (refs[3][0][0], refs[3][0][1]), (255, 0, 0)) # Z axis ref
+        return True
+    
+    def get_2d_pose(self, r):
+        top_center = r[0][0] + (r[1][0]-r[0][0])/2.0
+        bot_center = r[2][0] + (r[3][0]-r[2][0])/2.0
+        center = bot_center + (top_center - bot_center)/2.0
+        angle = np.arctan( (top_center[1]-bot_center[1]) / (top_center[0] - bot_center[0]) )
```
mattlangford: Check if the denom is zero before doing this

perception/sub8_perception/nodes/follow_orange_pipes.py
```diff
@@ -4,315 +4,228 @@
...
+        return True
+
+    def valid_contour(self, contour):
+        polygon = cv2.approxPolyDP(contour, 10, True)
+        if len(polygon) != 4:
+            return False
+        rect = self.sortRect(polygon)
+        for idx, p in enumerate(rect):
+            cv2.putText(self.last_image, str(idx), (p[0][0], p[0][1]), cv2.FONT_HERSHEY_SCRIPT_COMPLEX,1, (0,0,255))
+        length_width_ratio = np.linalg.norm(rect[0][0]-rect[3][0])/np.linalg.norm(rect[0][0]-rect[1][0])
```
mattlangford: Zero check the denominator here as well, something above could be ensuring that but unless you're sure of that, it's best to check

perception/sub8_perception/nodes/follow_orange_pipes.py
```diff
@@ -4,315 +4,228 @@
...
+        if abs(length_width_ratio-8.0)/8.0 > 0.2:
```
mattlangford: Can you disambiguate the randomish numbers

perception/sub8_perception/nodes/follow_orange_pipes.py
```diff
@@ -4,315 +4,228 @@
...
+        if not self.get_2d_pose(rect):
+            return False
+        if not self.get_3d_pose(rect):
+            return False
+        return True
+
+    def get_edges(self):
+        blur = cv2.blur(self.last_image, (5,5))
+        hsv = cv2.cvtColor(blur, cv2.COLOR_BGR2HSV)
+        thresh = cv2.inRange(hsv, (0, 100, 0), (60, 255, 255))
```
mattlangford: These should be params since it's almost certain that they won't be the same on a day to day and city to city basis. In the past there were sets of tuned numbers that you could use in the launch file so we could launch with pool numbers, sim numbers, or transdec numbers.

perception/sub8_perception/nodes/follow_orange_pipes.py
```diff
@@ -4,315 +4,228 @@
...
+    def img_cb(self, img):
+        if not self.enabled:
+            return
+        self.last_image = img
+        edges = self.get_edges()
+        _, contours, hierarchy = cv2.findContours(edges,cv2.RETR_TREE,cv2.CHAIN_APPROX_SIMPLE)
+        contours_img = img.copy()
+        for idx, c in enumerate(contours):
+            area = cv2.contourArea(c)
+            if area > 100:
```
mattlangford: The area will be dependent on the height from bottom. The mission should enforce a search height and the area should be computed using that.

perception/sub8_perception/nodes/test_path_marker.py
```diff
@@ -0,0 +1,104 @@
...
+        theta = np.arctan( (float(pts[1][1])-pts[0][1]) / (pts[1][0] - pts[0][0]) )
+        res.pose.x = middle[0]
+        res.pose.y = middle[1]
+        res.pose.theta = theta
+        return res
+
+    def _test_bag(self, filename, duration, pts):
+        rospy.sleep(rospy.Duration(0.5))
+        correct = self.get_ideal_response(pts)
+        print "Correct pose: {}".format(correct.pose.theta/(2*np.pi)*360.0)
```
mattlangford: `np.degrees` would be less messy, and it's more of a correct angle right?

perception/sub8_perception/nodes/follow_orange_pipes.py
```diff
@@ -4,315 +4,228 @@
...
+  - area (somewhat large)
+  - rectangular demensions (in rulebook, only check roughly)
+* cache response pose
+  - x,y from centroid (sub8 tools)
+  - middle of two skinny sides with numpy
+  - theta with atan of two skinny sides
+  - 
+
+* buffer / tracking / kalman filter
+  - so one bad frame doesn't fuck us
```
mattlangford: Kalman filter would be a cool idea. Can you somehow quantify your confidence in the observation?

perception/sub8_perception/nodes/follow_orange_pipes.py
```diff
@@ -4,315 +4,228 @@
...
+            self.markerPub = rospy.Publisher('path_marker_visualization', Marker, queue_size=10)
+        self.enabled = True
+
+    def sendDebugMarker(self):
+        m = Marker()
+        m.header.frame_id = self.image_sub.camera_info.header.frame_id
+        m.header.stamp = self.last_found_time
+        m.ns = "path_markers"
+        m.id = 0
+        m.type = 1 # sphere
```
mattlangford: You should use a scaled cube so you can see the predicted direction

command/sub8_missions/missions/align_channel.py
```diff
@@ -1,63 +1,52 @@
...
-    try:
-        resp = yield s.start_search(loop=False, timeout=600)
-    except:
-        print "MARKER_MISSION - Timed out probably."
-        defer.returnValue(None)
+    print "running pattern"
+    resp = yield s.start_search(loop=False, timeout=60)
 
-    if resp is None:
+    if resp == None:
```
mattlangford: _Techincally_ pep8 calls for `is None` over `== None`

perception/sub8_perception/nodes/follow_orange_pipes.py
```diff
@@ -4,315 +4,228 @@
...
+        sorted_x = np.array(sorted(rect, key=sort_contour_x))
+        
+        horizontal = False
+        if (np.linalg.norm(sorted_y[0] - sorted_y[1]) > np.linalg.norm(sorted_y[0] - sorted_y[2]) or
+                np.linalg.norm(sorted_y[0] - sorted_y[1]) > np.linalg.norm(sorted_y[0] - sorted_y[3])):
+            horizontal = True
+
+        sorted_rect = None
+        if horizontal:
+            if sorted_x[1][0][1] > sorted_x[0][0][1]:
```
mattlangford: Couldn't this all go up in the if statement above?

perception/sub8_perception/nodes/follow_orange_pipes.py
```diff
@@ -4,315 +4,228 @@
...
-
+            res.pose.x = self.last2d[0][0]
+            res.pose.y = self.last2d[0][1]
+            res.pose.theta = self.last2d[1]
+            res.found = True
+        return res
+
+    def get_3d_pose(self, p):
+        i_points = np.array((p[0][0], p[1][0], p[2][0], p[3][0]),dtype=np.float)
+        retval, rvec, tvec =  cv2.solvePnP(PathMarkerFinder.PATH_MARKER, i_points, self.cam.intrinsicMatrix(), np.zeros((5,1)))
```
mattlangford: I'd imagine this is not very robust to noisy data. It might be good to get the height off bottom and compare with that as a sanity check as well

perception/sub8_perception/nodes/follow_orange_pipes.py
```diff
@@ -4,315 +4,228 @@
...
-        self.image_sub = sub8_ros_tools.Image_Subscriber('/down/left/image_raw', self.image_cb)
-        self.image_pub = sub8_ros_tools.Image_Publisher("vision/channel_marker/target_info")
-
-        self.toggle = rospy.Service('vision/channel_marker/search', SetBool, self.toggle_search)
-
+        self.debug_ros = True
+        self.debug_gui = False
+        self.last2d = None
+        self.last3d = None
+        self.canny_low = 100
```
mattlangford: Parms

perception/sub8_perception/nodes/test_path_marker.py
```diff
@@ -0,0 +1,104 @@
...
+        # 10 pixel error accepted
+        res_xy = np.array([res.pose.x, res.pose.y])
+        correct_xy = np.array([correct.pose.x, correct.pose.y])
+        err = np.linalg.norm(res_xy-correct_xy)
+        msg="Marker pose (x,y) too much error Res={} Correct={} Error={}".format(res_xy, correct_xy, err)
+        self.assertLess(err, 
+                        10.0,
+                        msg=msg)
+        #5 degrees error
+        theta_err = abs(res.pose.theta-correct.pose.theta)
```
mattlangford: You can't subtract angles like this in the general case. Make sure the possible angles here are bounded in [-pi/2, pi/2] for this operation to be valid.

perception/sub8_perception/nodes/test_path_marker.py
```diff
@@ -0,0 +1,104 @@
...
+        msg = "Marker pose angle (theta) too much error Res={} Correct={} Error={}".format(np.degrees(res.pose.theta),
```
mattlangford: Is there weird spacing here or is it just my browser?

mattlangford (changes requested):
I'm guessing by the tests that the algorithm works, nice job. Adding a kalman filter to filter the pose of the marker would be a cool addition that would be good to learn how to do.

I think you need to add comments before merging and there are a few small other things in my comments.

sentree:
I asked @duggerd to create a continuous_testing directory on the SMB server. This will be used to store bags from the background Gazebo tests when that system is implemented. I could create a directory in that share for serving this data on our CI site, but I would like confirmation from @duggerd first to make sure he is fine with this solution.

sentree:
How does this react when there are multiple visible path markers?

perception/sub8_perception/nodes/follow_orange_pipes.py
kev-the-dev: u smart

perception/sub8_perception/nodes/follow_orange_pipes.py
kev-the-dev: So far if the contour is correct, the 3d translation is quite good. I use the 2D angle for orientation, but x,y,z have been quite accurate based on bags

kev-the-dev:
Currently, given a frame with multiple detected markers it would just return one of them at random. In my next iteration when I introduce a kalman filter I will allow for the case of two markers.

mattlangford:
Is multiple markers something we will have to deal with?

perception/sub8_perception/nodes/follow_orange_pipes.py
mattlangford: The key there being a correct contour. Using competition bags?

kev-the-dev:
I have seen one bag from competition where two markers are visible in a frame (I'll post a screenshot if I can find which one). Sounds like something AUVSI might do, so the perception should handle it.

perception/sub8_perception/nodes/follow_orange_pipes.py
kev-the-dev: Yes. Every competition bag I found showing path markers completely in frame for more than a second are included in the unit tests and this node passes the 2D test. I don't know the actual 3D pose to test to but the numbers seem reasonable (Certainly correct sign for x,y,z translation and in the ballpark)

kev-the-dev:
I believe all comments have been addressed -> trigger rereview

command/sub8_launch/launch/subsystems/blueview.launch
```diff
@@ -0,0 +1,21 @@
+<launch>
+    <node pkg="blueview_driver" type="blueview_driver" name="blueview_driver" >
+        <rosparam>
+            frame_id: blueview
+            device: 192.168.37.53
```
DSsoto: Would it be better if this was the dns name in case the ip changes? @whispercoros thoughts?

command/sub8_launch/launch/subsystems/cameras.launch
```diff
@@ -1,4 +1,4 @@
 <launch>
-  <include file="$(find sub8_launch)/launch/subsystems/cameras/launch/pt_grey_stereo.launch" />
-  <include file="$(find sub8_launch)/launch/subsystems/cameras/launch/downward.launch" />
+  <include file="$(find sub8_launch)/launch/subsystems/cameras/launch/front_cameras.launch" />
+  <include file="$(find sub8_launch)/launch/subsystems/cameras/launch/down_cameras.launch" />
```
DSsoto, line 3: You saw both front cameras run fine with the new launch?

command/sub8_launch/launch/tf.launch
DSsoto: I agree here tf2, is much nicer, the migration is pretty easy to execute too.
You switch from tf to tf2_ros and instantiate a tf2_ros.TransformListener() like this 
> `listener = tf2_ros.TransformListener(tf2_ros.Buffer())`.

This line would become: 
> `<node pkg="tf2_ros" type="static_transform_publisher" name="broadcater_name"
       args="0 0 0 0 0 0 1 parent_frame child_frame" />`

Notice that there is no frequency or period argument because it is a latched publisher.

command/sub8_missions/missions/align_channel.py
```diff
@@ -1,63 +1,48 @@
...
 def run(sub):
-    # Go to max height to see as much as we can.
-    yield sub.move.depth(SEARCH_DEPTH).zero_roll_and_pitch().go(speed=SPEED)
-    yield sub._node_handle.sleep(1.0)
-
-    # Do a little jig - change this for the pool.
-    pattern = [sub.move.right(1), sub.move.forward(1), sub.move.left(1), sub.move.backward(1),
-               sub.move.right(2), sub.move.forward(2), sub.move.left(2), sub.move.backward(2)]
-    s = Searcher(sub, sub.channel_marker.get_pose, pattern)
+    print "MARKER MISSION - STARTING"
```
DSsoto: Can you name this something like path alignment, or something related to alignment?
Just marker feels too ambiguous.

command/sub8_missions/sub8/sub_singleton.py
```diff
@@ -158,11 +159,11 @@ def __init__(self, node_handle):
 
     @util.cancellableInlineCallbacks
     def _init(self):
-        self._moveto_action_client = yield action.ActionClient(self.nh, 'moveto', MoveToAction)
-        self._odom_sub = yield self.nh.subscribe('odom', Odometry)
-        self._trajectory_sub = yield self.nh.subscribe('trajectory', PoseTwistStamped)
-        self._trajectory_pub = yield self.nh.advertise('trajectory', PoseTwistStamped)
-        self._dvl_range_sub = yield self.nh.subscribe('dvl/range', Float64Stamped)
+        self._moveto_action_client = yield action.ActionClient(self.nh, '/moveto', MoveToAction)
```
DSsoto: Whats the reason for this? Doesn't this just make the code less robust  to namespace changes?

For example, we run another small sub using the same stack and push all of these topic names down under the robot nam. I think this is the purpose of using relative names for topics.

perception/sub8_perception/CMakeLists.txt
```diff
@@ -202,3 +202,32 @@ target_link_libraries(
 install(
   TARGETS
 )
+
```
DSsoto: Can you put this test cmake code in a CMakeLists.txt in the test dir and just include it here using [`add_subdirectory(<path_to_testing_cmakelists>)`](https://www.johnlamp.net/cmake-tutorial-4-libraries-and-subdirectories.html#section-TestingForReal)?

This is likely to get a lot larger and I'd like to be able to quickly skim thorough the programs that are getting compiled here in this CMakeLists.

perception/sub8_perception/nodes/follow_orange_pipes.py
```diff
@@ -1,318 +1,269 @@
...
+    and returns this estimate when service is called.
+    
+    Unit tests for this node is in test_path_marker.py
+    
+    Finding the marker works as follows:
+    * blur image
+    * threshold image mostly for highly saturated, orange/yellow/red objects
+    * run canny edge detection on thresholded image
+    * find contours on edge frame
+    * filter contours to find those that may be contours by:
```
DSsoto: This seems to work great so it's fine, but in the future, and for more complex shapes, considered using [`cv2.matchShapes()`](http://docs.opencv.org/3.1.0/d5/d45/tutorial_py_contours_more_functions.html) which uses the Hu moment Invariants and should be much more robust than ad-hoc matching heuristics you come up with.

perception/sub8_perception/nodes/follow_orange_pipes.py
```diff
@@ -1,318 +1,269 @@
...
+      * checking # of sides in appox polygon
+      * checking ratio of length/width close to known model
+    * estimates 3D pose using cv2.solvePnP with known object demensions and camera model
+    
+    TODO: Implement Kalman filter to smooth pose estimate / eliminate outliers
+    """
+    # Model of four corners of path marker, centered around 0 in meters
+    PATH_MARKER = np.array([[0.6096,  -0.0762, 0],
```
DSsoto: This should be a ros param. You could also replace the `marker.scale.*` magic constants in the rviz pub function with an expression using it.

perception/sub8_perception/nodes/test_path_marker.py
```diff
@@ -0,0 +1,122 @@
+#!/usr/bin/env python
```
DSsoto: This file should go in the test directory.

perception/sub8_perception/nodes/test_path_marker.py
```diff
@@ -0,0 +1,122 @@
...
+    #Disabled because not long enough
+    #~ def test_transdec2(self):
+        #~ self._test_bag('/home/kallen/bag/path_marker_transdcec/fixed/transdec_path2.bag',
+                       #~ rospy.Duration(1.9),
+                       #~ [ [392, 166], [428,16] ])
+    def test_transdec3(self):
+        self._test_bag(rospy.get_param(self.name+'/transdec_path3'),
+                       rospy.Duration(0.8),
+                       [ [244,9],[321, 127] ])
+    # Disabled because too noisey
```
DSsoto: lol if it fails it's bullshit! +1

utils/sub8_ros_tools/sub8_ros_tools/fix_bag.py
```diff
@@ -0,0 +1,78 @@
+#!/usr/bin/env python
```
DSsoto: Good stuff!

DSsoto (changes requested):
Requesting a couple of changes and have a couple of comments I'd like addressed. Would like to merge today if possible.

command/sub8_launch/launch/tf.launch
```diff
@@ -29,6 +29,10 @@
 
     <node pkg="sub8_launch" type="tf_republisher.py" name="tf_replublisher" />
 
+    <node pkg="tf" type="static_transform_publisher" name="blueview_broadcaster"
+      args="0.333 0 0.381  0 0 0 1 base_link blueview 100">
```
RustyBamboo: I think the blueview orientation is <0 0 -0.707 0.707> relative to base_link, where the x-axis is pointing right, y-axis is forward, and z-axis is up.

mattlangford: Is that the case? I thought the sonar was pointing straight forward?

command/sub8_missions/missions/align_channel.py
mattlangford: Yeah, I didn't like all the mixed references to marker/orange_path/alignment/etc last year

Also you should use fprint for the colors and headers (since there's a header in every print line)

perception/sub8_perception/nodes/follow_orange_pipes.py
mattlangford: Does it need to be? It won't change will it? It's better not to clutter up the things that we could change with things that never would.

perception/sub8_perception/nodes/follow_orange_pipes.py
mattlangford: That brings up a good point of non-fully in frame markers, do you handle that? And only on 2 transdec bags?

utils/sub8_ros_tools/sub8_ros_tools/fix_bag.py
```diff
@@ -0,0 +1,78 @@
...
+  }
+
+  def fix_topic(self, topic):
+      for k, t in BagFixer.TOPIC_MAP.iteritems():
+          if topic.find(k) == 0:
+            return t+topic[len(k):]
+      return topic
+
+  def fix_tf(self, tf):
+    for i, t in enumerate(tf.transforms):
```
mattlangford: I see your two spaces, don't think I've gone blind in my old age.

utils/sub8_ros_tools/sub8_ros_tools/fix_bag.py
```diff
@@ -0,0 +1,78 @@
...
+    self.inbag = inbag
+    if outbag == 'DEFAULT':
+      split = inbag.rsplit('.bag', 1)
+      outbag = split[0]+'_fixed.bag'
+    self.outbag = outbag
+    self.fix_bag()
+
+
+
+if __name__ == "__main__":
```
mattlangford: This could be a useful general tool, maybe add an option to specify `from` and `to` topics where you can call it like: `fix_bag.py --in dumb_topic_name --out good_topic_name --in another_bad_topic --out good_topic_2`. That would fix topics but you'd have to specify framing in a similar way. This would complicate the thing this file is doing right now but make it a lot more general.

mattlangford (approved):
I'm cool with this. I think any changes on the base algorithm (kalman filter, more robust detection, knowing if the thing isn't fully in frame) could go in another PR. Defering to David for final verdict.

I would like to see fprint used in the mission.

command/sub8_launch/launch/tf.launch
kev-the-dev: It is, but the data it produces treats X, Y, and Z as Daniel described. So having the correct transform here would be nice so the ogrid/pointcloud published from the data will transform correctly.

command/sub8_launch/launch/tf.launch
kev-the-dev: Let's not switch to tf2 in the PR. Seems like it might offer some benefits, but not sure what it will break. Specifically, I don't know if txros supports it, which could break missions.

command/sub8_missions/sub8/sub_singleton.py
kev-the-dev: I changed it because I was having missions get stuck waiting for odom forever in sim, doesn't seem to be happening any more... huh. I'll switch it back

perception/sub8_perception/nodes/follow_orange_pipes.py
kev-the-dev: The notes.txt for the transdec bags says only 3 have path markers, and one was half out of frame :(. If it's out of frame, I don't detect it

command/sub8_launch/launch/subsystems/blueview.launch
kev-the-dev: Don't know the domain and not sure if the sdk handles domains, but I'm happy to try it.

perception/sub8_perception/nodes/test_path_marker.py
kev-the-dev: Okay, I moved it. It might cause issues though because it is not in one of the two default folders for python nodes. Do I need to update setup.py?

perception/sub8_perception/nodes/test_path_marker.py
kev-the-dev: It's so glarry I can barely identify it myself. Hope to get a few more testing bags once new down camera is installed.

utils/sub8_ros_tools/sub8_ros_tools/fix_bag.py
kev-the-dev: Good idea, I made it a generic tool and added a [wiki page](https://github.com/uf-mil/SubjuGator/wiki/Working-with-old-bags) for using it with subjugator bags. I'll probably move it to mil_common once it has a tools package in master branch.

perception/sub8_perception/nodes/test_path_marker.py
DSsoto: Nothing should really need to import from this right? And yea, if anything breaks, add an __init__.py and update the setup.py.

perception/sub8_perception/nodes/follow_orange_pipes.py
DSsoto: I kind of like the idea of each vision node having a yaml where all it's relevant parameters are defined. That way if anything needs to be changed, everyone knows where to go. No need to skim through all the source code.

I'll leave it up to Kevin's discretion.

command/sub8_launch/launch/tf.launch
DSsoto: Sure, but after this, I will be very hesitant to merge stuff not using tf2. Tf has been deprecated since late 2013 (Hydro) and is pretty buggy.

DSsoto (approved):

command/sub8_launch/launch/subsystems/blueview.launch
DSsoto: mil-com-teledyne-p900.ad.mil.ufl.edu

command/sub8_launch/launch/subsystems/cameras.launch
kev-the-dev: Yes. Both looked good besides green bug and had correct frames in camera info and image topics

kev-the-dev: Also going to disable down camera in launch because it is not connected.

kev-the-dev:
Updated to use DNS for blueview. Good to go!

mattlangford:
Someone should try tf2 with txros, and make a PR to support it if it doesn't work

PR #212, opened 2017-04-10
START_GATE: Removed unnecessary imports

RustyBamboo:


command/sub8_missions/missions/start_gate.py
```diff
@@ -7,7 +7,6 @@
 import time
```
DSsoto: It seems like you also don't need rospkg, or time.

You should replace sleeping with `time` to sleeping with `txros.sleep()`.

DSsoto (changes requested):
Just one thing.

DSsoto (approved):

PR #219, opened 2017-04-13
CI: Integrate SubjuGator with the Jenkins CI server

sentree:
This is the first repository to be integrated with the new CI stack. The server uses Jenkins with a Docker backend to ensure that all builds are done in a reproducible environment. All pull requests and branches will be built based on the Jenkins file in the root directory, so that can be changed if we need to add a stage.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,0 +1,25 @@
+dockerNode(image: 'uf-mil:subjugator') {
+	stage("Checkout") {
+		checkout scm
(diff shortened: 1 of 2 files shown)


DSsoto (approved):
I don't really understand this but I have seen it work on @whispercoros' machine.

sentree:
Thanks to @ironmig for figuring out the "checkout scm" line.

PR #222, opened 2017-04-16
CI: fail Jenkins build if tests dont pass

kev-the-dev:
* I think this will make the Jenkins build fail of a unit test fails. Also will display good summary of what tests passed 

* Tested: does FAIL when a test fails, see build PR-222 Run 2

* Tested: still PASSES when all tests pass, see build PR-222 Run 3


DSsoto:
Can we get colored output from Jenkins?

DSsoto (approved):

sentree:
This change was well tested, but let me review everything that goes to CI before merging, please.

whispercoros:
This change was well tested, but let me review everything that goes to CI before merging, please.

sentree:
I will add colored output to the list, but there are more important things to do for CI right now.

PR #223, opened 2017-04-16
Thrusters, Sim, Diagnostics, Utils

DSsoto:
* Unifies thruster configuration for both simulated and real thrusters (removes a lot of outdated config files)
* Adds a utility to quickly conduct thruster spin tests
* Adds checking if the front cameras are connected to the usb bus (self check)
* Make the self check specifically list which thrusters are missing


DSsoto:
Before anyone calls me out on it, I see that I've broken the `test_thruster_comm` test. I will fix before merge. Sweet, sweet CI !!

kev-the-dev:
If you merge #222 would get rid of that little check mark for ya ;)

command/sub8_launch/nodes/self_check.py
```diff
@@ -136,6 +146,20 @@ def tx_init(self):
     @txros.util.cancellableInlineCallbacks
     def do_check(self):
         passed = True
+
+        # Check if front cameras are actally on usb bus
+        command = "lsusb -d {}".format(self.front_cam_product_id)
+        err_str = "{} front camera{} not connected to usb port"
+        try:
+            count_front_cam_usb = \
+                subprocess.check_output(["/bin/sh", "-c", command]).count("Point Grey")
```
mattlangford, line 155: Do you have to have the first two commands here? Won't just calling lsusb -d work?

drivers/sub8_videoray_m5_thruster/nodes/thruster_spin_test.py
```diff
@@ -0,0 +1,90 @@
+#!/usr/bin/env python
+
+if __name__ == '__main__':
+    import sys
```
mattlangford: Why are the imports below the check for main

drivers/sub8_videoray_m5_thruster/nodes/thruster_spin_test.py
```diff
@@ -0,0 +1,90 @@
...
+
+    def command_thrusters(timer_event):
+        for motor_id in active_thrusters:
+            name = names_from_motor_id[motor_id]
+            thruster_ports[name].send_thrust_msg(motor_id, thrust)
+
+    timer = rospy.Timer(period=rospy.Duration(0.01), callback=command_thrusters)
+
+    while not rospy.is_shutdown():
+       rospy.Rate(20) # just enough to be interesting
```
mattlangford: Does this actually sleep? I thought you had to call `sleep` or something

drivers/sub8_videoray_m5_thruster/nodes/thruster_spin_test.py
mattlangford: And everything for that matter.

This could be something that might be easier as an rqt plugin if Anthony is bent on making on

drivers/sub8_videoray_m5_thruster/nodes/thruster_spin_test.py
```diff
@@ -0,0 +1,90 @@
...
+    '''
+    fprint(usage_msg)
+
+    thrust = 0.5  # Normalized thrust in [0, 1]
+    key = None
+
+    def command_thrusters(timer_event):
+        for motor_id in active_thrusters:
+            name = names_from_motor_id[motor_id]
+            thruster_ports[name].send_thrust_msg(motor_id, thrust)
```
mattlangford: Maybe you could constrain the thrust here (`np.clip(thrust, 0.0, 1.0)`) instead of doing it each time you update.

drivers/sub8_videoray_m5_thruster/nodes/thruster_spin_test.py
```diff
@@ -0,0 +1,90 @@
...
+                           4: 'BLH', 5: 'BLV', 6: 'BRH', 7: 'BRV'}
+    active_thrusters = set()
+
+    fprint = FprintFactory(title='thruster_spin_test', time=rospy.Time.now, auto_bold=False).fprint
+    usage_msg = \
+    '''
+    Welcome to the thruster_spin_test tool.
+    Instructions:
+    * press up or down to control the thrust commanded
+    * press the thruster id to spin that specific thruster (1-indexed, 0 means all)
```
mattlangford: Would it be better to trigger them by name rather than ID

drivers/sub8_videoray_m5_thruster/sub8_thruster_comm/thruster_comm.py
```diff
@@ -68,15 +77,18 @@ def connect_port(self, port_name):
             raise(e)
         return serial_port
 
-    def load_thruster_config(self, thruster_name, thruster_info):
+    def load_thruster_config(self, thruster_name, thruster_definitions):
+        '''Loads the motor_id from the config file if the thruster was found'''
         # Get our humungous error string ready
         errstr = "Could not find motor_id {} (Called {}) on port {}; existing ids are {}".format(
             thruster_info['node_id'], thruster_name, self.port_name, self.motor_ids_on_port)
```
mattlangford: Does thruster info exist anymore?

simulation/sub8_gazebo/src/sub8_thruster_config.cpp
```diff
@@ -0,0 +1,100 @@
...
+  }
+
+  for(int i = 0; i < ports_xmlrpc.size(); ++i)  // Iterate through list of port defs
+  {
+    auto port_def = ports_xmlrpc[i];
+    auto port = ThrusterPort();
+    auto names = port_def["thruster_names"];
+
+    port.port_name = (string)port_def["port"];
+    for(int i = 0; i < names.size(); ++i)
```
mattlangford: Redefinition of `i`

mattlangford (commented):
Video Ray does make a distinction between `motor_id` and`node_id`. I don't remember which we actually used to address each thruster, but if it's `node_id`, probably best to leave it like it was.
Approval pending comment responses.

drivers/sub8_videoray_m5_thruster/nodes/thruster_spin_test.py
DSsoto: Fixed/.

It could be, but it's pretty nice and easy to use as it is. Not a good use of his time.

drivers/sub8_videoray_m5_thruster/nodes/thruster_spin_test.py
DSsoto: No because it's only one key press for each thruster so you can easily toggle each thruster. Plus they're all labeled anyways.

drivers/sub8_videoray_m5_thruster/nodes/thruster_spin_test.py
DSsoto: I did it down there so that I could display the new thrust value whenever it is changed.

drivers/sub8_videoray_m5_thruster/nodes/thruster_spin_test.py
DSsoto: Fixed

simulation/sub8_gazebo/src/sub8_thruster_config.cpp
DSsoto: Good catch, fixed.

mattlangford (approved):
Did you roll with this last test day and see that everything is good?

Also again with the naming, Video Ray does make the distinction between node ID and motor ID - can you get with @duggerd and confirm we actually address them by motor ID?

Also I don't know if I should wait for CI to merge or something so I'll standby.

DSsoto:
We ran this version of the repo at the pool and everything worked before thruster slip became an issue. I didn't manage to get the spin test utility running by test day, but I got it working last night so I can confirm that I can spin all of the thrusters by motor_id.

I just fixed the broken rostest, so this is ready to merge.

PR #224, opened 2017-04-17
SCRIPTS: Account for changes in mil_common

sentree:
This pull request maintains compatibility between the SubjuGator and mil_common scripts. A few bagging aliases were added for the GNC topics and the Jenkins file now supports any submodules we may add at a later date.


scripts/bagging_variables.sh
```diff
@@ -22,4 +22,7 @@ export bag_blueview_ranges="/blueview_driver/ranges"
 export bag_blueview_color="/blueview_driver/image_color"
 export bag_blueview_mono="/bluewview_driver/image_mono"
```
kev-the-dev: Just noticed this typo (my b), can you fix it in this PR?

scripts/bash_aliases.sh
```diff
@@ -8,7 +8,7 @@
 
 
 # Directory navigation
-alias sub="cd $CATKIN_DIR/src/SubjuGator"
+alias sub="cd \$CATKIN_DIR/src/SubjuGator"
```
kev-the-dev, line 11: Again I encounter the mysterious backslash. Are we on windows or something?

kev-the-dev (commented):
Good Juice

scripts/bash_aliases.sh
sentree: The change escapes the `$` character, which means the variable's value will be based on when the alias is used rather than when it is set.

scripts/bagging_variables.sh
sentree: This is fixed.

DSsoto (approved):
Approve

PR #225, opened 2017-04-18
BUGS: Fix problems with older compilers

kingkevlar:
I was unable to compile the repo on my systems after a fresh install script using Debian 8.7

* Fix eigen includes
* Fix thruster string type ambiguity

### DIFF
--- a/gnc/.sub8_trajectory_generator/CMakeLists.txt
+++ b/gnc/.sub8_trajectory_generator/CMakeLists.txt
@@ -6,7 +6,7 @@ project(${PROJECT_NAME})
 find_package(catkin REQUIRED COMPONENTS roscpp geometry_msgs sub8_msgs cmake_modules)
 find_package(OMPL REQUIRED)
 find_package(Boost REQUIRED)
-find_package(Eigen REQUIRED)
+find_package(EIGEN REQUIRED)
 # find_package(sub8_alarm REQUIRED)
 
 set(catkin_LIBRARIES
@@ -29,7 +29,7 @@ include_directories(
  	${catkin_INCLUDE_DIRS}
   	${OMPL_INCLUDE_DIRS}
   	${Boost_INCLUDE_DIRS}
-  	${Eigen_INCLUDE_DIRS}
+  	${EIGEN_INCLUDE_DIRS}
   	${sub8_alarm_INCLUDE_DIRS}
 )
 
@@ -44,7 +44,7 @@ include_directories(
 #target_link_libraries(${PROJECT_NAME}
 #	${catkin_LIBRARIES}
 #	${OMPL_LIBRARIES}
-#	${Eigen_LIBRARIES}
+#	${EIGEN_LIBRARIES}
 #	${sub8_alarm_LIBRARIES}
 #)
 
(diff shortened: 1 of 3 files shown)


sentree:
Thanks for doing this!

RustyBamboo:
Could an older version of GCC (GCC 4.9 on Debian) also be causing problems?

sentree:
We are on GCC 4.9.2. It may be a good idea to standardize the compiler version since it seems backwards compatibility may be an issue, but this is the only problem we have ever encountered. This solution compiles with both versions, so I'm fine just merging the changes for now.

simulation/sub8_gazebo/src/sub8_thruster_config.cpp
```diff
@@ -35,7 +35,7 @@ try
 catch(XmlRpcException& e)
 {
   auto err_str = "Error parsing port definitions: %s";
```
DSsoto, line 37: @Hmmm its interesting, on g++ 5.4 ( the default for Ubuntu 16.04) err_str is already a c-string so the conversion below is not needed. @kingkevlar can you post the output of `g++ --vesion`?

DSsoto:
G++ 4.9 is the minimum required for ROS Kinetic so maybe that's the one that the CI should compile with.

DSsoto (commented):
This compiled fine on my machine ( g++ 5.4).

I would approve but since you can look at [this](https://github.com/uf-mil/SubjuGator/search?utf8=%E2%9C%93&q=Eigen%2FCore&type=) and see that all of these other ocurrences of similar includes didn't require you to change them, this is not really the root cause of the issue.

On my machine, `$(EIGEN_INCLUDE_DIRS)` adds `/usr/include/eigen3/` to the include path. Inside that directory is `Eigen/Core`, `Eigen/Dense` and everything else we use.

simulation/sub8_gazebo/src/sub8_thruster_config.cpp
kingkevlar: 4.9.2

kev-the-dev: The err_str is not converted to c string, the XMLRpc exception call (which returns std::string) is.

DSsoto: Yup, missed that. Although err_str is interpreted by the compiler to be a char ptr.

kev-the-dev (approved):
Fixes build on some systems and doesn't break CI. I'm happy

gnc/.sub8_trajectory_generator/CMakeLists.txt
```diff
@@ -6,7 +6,7 @@ project(${PROJECT_NAME})
 find_package(catkin REQUIRED COMPONENTS roscpp geometry_msgs sub8_msgs cmake_modules)
 find_package(OMPL REQUIRED)
 find_package(Boost REQUIRED)
-find_package(Eigen REQUIRED)
+find_package(EIGEN REQUIRED)
```
DSsoto, line 9: It doesn't matter since this is not currently being compiled, but this was supposed to be `Eigen3`.

PR #228, opened 2017-04-24
ONLINE_BAGGER: Add to sub launch and kill systems

kingkevlar:
Oooooweeeee! Online_bagger has been added to the launch and kill systems for the sub. (launches with the sub and calls the service dump on kill)

Whenever 'araise kill' is called, the current message set is dumped to a bag called 'kill_bag' (see PR in mil_common). It will repeatedly get written over with the name 'kill_bag.bag' to prevent having too many kill bags (this makes it our responsibility to rename bags we care about). I can switch back to the default bag name (current date and time) if we reach some sort of consensus on it. 

I also updated the yaml to bag one camera (right) and the blueview images and ranges by default. (Again, can be changed)

### DIFF
--- a/command/sub8_alarm/alarm_handlers/kill.py
+++ b/command/sub8_alarm/alarm_handlers/kill.py
@@ -1,5 +1,6 @@
 import rospy
 from ros_alarms import HandlerBase
+from mil_msgs.srv import BaggerCommands
 
 class Kill(HandlerBase):
     alarm_name = 'kill'
@@ -8,12 +9,21 @@ class Kill(HandlerBase):
     def __init__(self):
         self._killed = False
         self._last_mission_killed = False
+        self.bagging_server = rospy.ServiceProxy('/online_bagger/dump', BaggerCommands)
 
     def raised(self, alarm):
         self._killed = True
+        self.bagger_dump()
 
     def cleared(self, alarm):
         self._killed = False
+
+    def bagger_dump(self):
+        """Call online_bagger/dump service"""
+        try:
+            bag_status = self.bagging_server(bag_name='kill_bag', bag_time=0)
+        except rospy.ServiceException, e:
+            print "/online_bagger service failed: %s" %e
 
     def meta_predicate(self, meta_alarm, sub_alarms):
         ignore = []
--- a/command/sub8_alarm/alarm_handlers/kill.py
+++ b/command/sub8_alarm/alarm_handlers/kill.py
@@ -20,8 +20,13 @@ class Kill(HandlerBase):
 
     def bagger_dump(self):
         """Call online_bagger/dump service"""
+        camera = '/camera/front/left/camera_info /camera/front/left/image_raw '
+        navigation = '/wrench /wrench_actual /c3_trajectory_generator/trajectory_v /c3_trajectory_generator/waypoint /trajectory '
+        controller = '/pd_out /rise_6dof/parameter_descriptions /rise_6dof/parameter_updates '
+        blueview = '/blueview_driver/ranges /blueview_driver/image_color'
+        kill_topics = camera + navigation + controller + blueview
         try:
-            bag_status = self.bagging_server(bag_name='kill_bag', bag_time=60)
+            bag_status = self.bagging_server(bag_name='kill_bag', bag_time=60, topics=kill_topics)
         except rospy.ServiceException, e:
             print "/online_bagger service failed: %s" %e
 
--- a/command/sub8_alarm/alarm_handlers/kill.py
+++ b/command/sub8_alarm/alarm_handlers/kill.py
@@ -21,7 +21,7 @@ class Kill(HandlerBase):
     def bagger_dump(self):
         """Call online_bagger/dump service"""
         try:
-            bag_status = self.bagging_server(bag_name='kill_bag', bag_time=0)
+            bag_status = self.bagging_server(bag_name='kill_bag', bag_time=60)
         except rospy.ServiceException, e:
             print "/online_bagger service failed: %s" %e
 
--- a/command/sub8_launch/config/online_bagger.yaml
+++ b/command/sub8_launch/config/online_bagger.yaml
@@ -14,3 +14,8 @@ stream_time: 60  # seconds
 
 # sets bag package path, default to home/user/bags/current-date
 # bag_package_path : "home/sub8/mil_ws/bags"
+
+# Bool parameter that determines if current date is appended to bag directory
+# for example defaults to: "home/user/bags/current-date" when True
+# else bag directory goes to: "home/user/bags when False
+dated_folder: True
--- a/command/sub8_launch/config/online_bagger.yaml
+++ b/command/sub8_launch/config/online_bagger.yaml
@@ -1,0 +1,15 @@
+# list an arbitrary set of topics to  subscribe to and stream:
+
+stream_time: 10  # seconds
+
+topics: [["/odom",                    300] ,
+         ["/absodom",                 300],
+         ["/stereo/left/camera_info"     ],
+         ["/stereo/left/image_raw",    20],
+         ["/stereo/right/camera_info", 20],
+         ["/stereo/right/image_raw",   20], 
+         ["/velodyne_points",         300],
+         ["/ooka",                    100]]
+
+# comment out if default home/user/online_bagger is desired
+# bag_package_path: ''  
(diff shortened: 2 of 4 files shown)


command/sub8_alarm/alarm_handlers/kill.py
```diff
@@ -8,13 +9,23 @@ class Kill(HandlerBase):
...
 
     def raised(self, alarm):
         self._killed = True
+        self.bagger_dump()
 
     def cleared(self, alarm):
         self._killed = False
 
+    def bagger_dump(self):
+        ''' Dump online_bagger'''
```
sentree: Try to keep your documentation strings in this format:

    def do_things(self):
    '''
    This function does things.
    ''''
    self.stuff()
    self.things()

Not that there are no spaces between the function definition, documentation string, or first line of code. The only reason for spacing here is if the first line after the documentation string is a comment. I believe this is PEP8, but it may just be OCD; someone correct me if I'm wrong.

command/sub8_alarm/alarm_handlers/kill.py
```diff
@@ -8,13 +9,23 @@ class Kill(HandlerBase):
...
+
+        try:
+            bag_status = self.bagging_server(bag_name='kill_bag', bag_time=0)
```
sentree: By passing bag_time=0, won't we be bagging everything in the cache? Not sure how I feel about this as we may want a larger queue later down the line, but not want to dump the whole thing for a kill.

sentree (approved):
I saw you and @ironmig freak out about this working, so it was likely well tested.

sentree:
I try to keep bags in a separate partition so they never interfere with our root or home partitions. Also, please make sure your bag ends up in the correct dated folder as defined in the mil_common bag.sh script. Or at least keep in the default bagging folder, which is stored in the `BAG_DIR` environment variable.

command/sub8_alarm/alarm_handlers/kill.py
kingkevlar: I believe PEP8 calls for docstrings in double quotes (which I did not do) and:

 PEP8:
"For one liner docstrings, please keep the closing """ on the same line. "

sentree: Okay, now I have a beef with PEP8... well, convention is convention!

command/sub8_alarm/alarm_handlers/kill.py
kingkevlar: I didn't think of that, do you think 60 seconds would be an appropriate time?

sentree (changes requested):
On hold until @ironmig's mil_common pull request is merged.

kev-the-dev:
I'm fine with merging once uf-mil/mil_common#48 is merged.

sentree:
Will this integrate well with all of the changes you made?

kev-the-dev:
Doesn't really interface with my changes. But I did pull this branch and test it again uf-mil/mil_common#48 and it seems fine.

command/sub8_launch/config/online_bagger.yaml
```diff
@@ -0,0 +1,12 @@
+# list an arbitrary set of topics to  subscribe to and stream:
+
+stream_time: 60  # seconds
+
+topics: [["/odom",                          ],
```
kev-the-dev: This list is unnecessary now that it pulls from the ```bag_*``` environment variables. The bagging topics will now be version controlled through the bag aliases file.

kev-the-dev (changes requested):
One teency redundancy we should get rid of now that onine bagger is integrated with the bag script.

command/sub8_alarm/alarm_handlers/kill.py
```diff
@@ -8,13 +9,22 @@ class Kill(HandlerBase):
...
         self._killed = True
+        self.bagger_dump()
 
     def cleared(self, alarm):
         self._killed = False
 
+    def bagger_dump(self):
+        """Call online_bagger/dump service"""
+        try:
+            bag_status = self.bagging_server(bag_name='kill_bag', bag_time=60)
```
sentree: Are we sure we want to dump all topics for the kill bag?

command/sub8_launch/launch/subsystems/online_bagger.launch
```diff
@@ -0,0 +1,4 @@
+<launch>
+    <node pkg="mil_tools" type="online_bagger.py" name="online_bagger" output="screen"/>
+    <rosparam file="$(find sub8_launch)/config/online_bagger.yaml" command="load" ns="online_bagger"/>
+</launch>
```
sentree: All of @ironmig's shiny new ROS parameters should be added to this launch file. Make sure to leave the bag storage directory blank so it defaults to using the `BAG_DIR` environment variable.

sentree (commented):
This is close to ready. It just needs to be updated a bit before it is meged.

command/sub8_alarm/alarm_handlers/kill.py
DSsoto: Hmmm, maybe it should only be the GNC ones and one camera / imaging sonar. Then again maybe we could spare 60 seconds of everything just for kill, not sure how large that gets.

sentree: Make sure to check next time you guys test. Launch this from @kingkevlar's workspace.

DSsoto:
@kingkevlar, @ironmig, what's the status on this?
I remember seeing you guys fix some bugs but nothing has been pushed here since.

command/sub8_alarm/alarm_handlers/kill.py
```diff
@@ -8,13 +9,27 @@ class Kill(HandlerBase):
...
 
+    def bagger_dump(self):
+        """Call online_bagger/dump service"""
+        camera = '/camera/front/left/camera_info /camera/front/left/image_raw '
+        navigation = '/wrench /wrench_actual /c3_trajectory_generator/trajectory_v /c3_trajectory_generator/waypoint /trajectory '
+        controller = '/pd_out /rise_6dof/parameter_descriptions /rise_6dof/parameter_updates '
+        blueview = '/blueview_driver/ranges /blueview_driver/image_color'
+        kill_topics = camera + navigation + controller + blueview
+        try:
+            bag_status = self.bagging_server(bag_name='kill_bag', bag_time=60, topics=kill_topics)
```
mattlangford, line 29: Have you tried having the sub kill unkill and kill again within a few seconds (or calling the service right after the sub killed)? It looks like there could be a race condition where both requests will try writing to different bags using the same bag object. That may get caught here as an exception but it'll still crash the bag server.

kingkevlar: Do you mean calling the same service simultaneously: as in, while it is being called by the alarm server, it is also getting called in another terminal separately? <- I haven't checked this case, I can check it tomorrow, I didn't think of it.

Or did you mean closely in series as in: run kill (dump bag), then run service again immediately after? <- This one does not cause issues

mattlangford: The former, or just trying to dump a bag while another is being dumped

DSsoto: @kingkevlar, If that's the case, putting a mutex around the `BaggingCommands` function should fix the issue.

kingkevlar: It didn't work when I tried to run them simultaneously, if it was in the middle of a service call, it seems that it would start running the second one in the middle of the first one and the resulting two bags would have the topics split between the two service calls (i.e. I had one bag that had one of the camera topics, and another than had the other half, both had the same amount of messages).

I fixed the issue with a PR to mil_common, if the service is already being called it will reject the call and give priority to the current one. (aka, you cannot run the service in parallel with another service call)

kingkevlar:
@whispercoros It says the commit cannot be built, when I looked at the details it looks like it is related to the thruster_mapper

kev-the-dev (approved):

command/sub8_alarm/alarm_handlers/kill.py
sentree, line 29: Everyone remember that we will need to rename the kill bag if we want to save it.

sentree (approved):
Everything looks good here. Nice job!

PR #229, opened 2017-05-02
PATH MARKER: Use kalman filter to reduce pose estimate noise

kev-the-dev:
* Use kalman filter for smoothing noise in path marker vision. Seems to work well in testing. Should ensure we don't accidentally try to move to some visual artifact that looks like a path marker
* Enable down camera in camera.launch, restore calibration file from a while back (might not be good)

### DIFF
--- a/command/sub8_missions/missions/align_path_marker.py
+++ b/command/sub8_missions/missions/align_path_marker.py
@@ -2,6 +2,7 @@ from __future__ import division
 
 import genpy
 import tf
+import numpy as np
 from twisted.internet import defer
 from txros import util
 from sub8 import Searcher
@@ -10,39 +11,57 @@ from mil_misc_tools import text_effects
 import numpy as np
 
 SEARCH_DEPTH = .65
+SEARCH_RADIUS_METERS = 1.0
 TIMEOUT_SECONDS = 60
+DO_PATTERN=True
+SPEED = 0.5
 MISSION="Align Path Marker"
+
+forward_vec = np.array([1, 0, 0, 0])
 @util.cancellableInlineCallbacks
 def run(sub):
-    print_info = text_effects.FprintFactory(title=MISSION)
-    print_bad = text_effects.FprintFactory(title=MISSION, msg_color="red")
-    print_good = text_effects.FprintFactory(title=MISSION, msg_color="green")
-    print_info.fprint("STARTING")
+    print_info = text_effects.FprintFactory(title=MISSION).fprint
+    print_bad = text_effects.FprintFactory(title=MISSION, msg_color="red").fprint
+    print_good = text_effects.FprintFactory(title=MISSION, msg_color="green").fprint
+    print_info("STARTING")
 
     # Wait for vision services, enable perception
-    print_info.fprint("ACTIVATING PERCEPTION SERVICE")
+    print_info("ACTIVATING PERCEPTION SERVICE")
     sub.vision_proxies.path_marker.start()
 
     pattern = []
-    #pattern = [sub.move.right(1), sub.move.forward(1), sub.move.left(1), sub.move.backward(1),
-     #          sub.move.right(2), sub.move.forward(2), sub.move.left(2), sub.move.backward(2)]
+    if DO_PATTERN:
+        pattern = [sub.move.right(SEARCH_RADIUS_METERS), sub.move.forward(SEARCH_RADIUS_METERS),
+                   sub.move.left(SEARCH_RADIUS_METERS), sub.move.backward(SEARCH_RADIUS_METERS),
+                   sub.move.right(2.0*SEARCH_RADIUS_METERS), sub.move.forward(2.0*SEARCH_RADIUS_METERS),
+                   sub.move.left(2.0*SEARCH_RADIUS_METERS), sub.move.backward(2.0*SEARCH_RADIUS_METERS)]
     s = Searcher(sub, sub.vision_proxies.path_marker.get_pose, pattern)
     resp = None
-    print_info.fprint("RUNNING SEARCH PATTERN")
+    print_info("RUNNING SEARCH PATTERN")
     resp = yield s.start_search(loop=False, timeout=TIMEOUT_SECONDS, spotings_req=1)
 
     if resp is None or not resp.found:
-        print_bad.fprint("MARKER NOT FOUND")
+        print_bad("MARKER NOT FOUND")
         defer.returnValue(None)
-    
-    print_good.fprint("PATH MARKER POSE FOUND")
+
+    print_good("PATH MARKER POSE FOUND")
     assert(resp.pose.header.frame_id == "/map")
-    move = sub.move.set_orientation(rosmsg_to_numpy(resp.pose.pose.orientation)).zero_roll_and_pitch()
+
+    move = sub.move
     position = rosmsg_to_numpy(resp.pose.pose.position)
-    position[2] = move._pose.position[2]
-    move = move.set_position(position)
-    print_info.fprint("MOVING TO MARKER POSE")
-    yield move.go(speed=0.2)
-    print_good.fprint("ALIGNED TO PATH MARKER, DONE!")
+    position[2] = move._pose.position[2] # Leave Z alone!
+    orientation = rosmsg_to_numpy(resp.pose.pose.orientation)
+
+    move = move.set_position(position).set_orientation(orientation).zero_roll_and_pitch()
+
+    # Ensure SubjuGator continues to face same general direction as before (doesn't go in opposite direction)
+    odom_forward = tf.transformations.quaternion_matrix(sub.move._pose.orientation).dot(forward_vec)
+    marker_forward = tf.transformations.quaternion_matrix(orientation).dot(forward_vec)
+    if np.sign(odom_forward[0]) != np.sign(marker_forward[0]):
+        move = move.yaw_right(np.pi)
+
+    print_info("MOVING TO MARKER AT {}".format(move._pose.position))
+    yield move.go(speed=SPEED)
+    print_good("ALIGNED TO PATH MARKER. MOVE FORWARD TO NEXT CHALLENGE!")
     sub.vision_proxies.path_marker.stop()
     defer.returnValue(True)
--- a/command/sub8_missions/missions/align_path_marker.py
+++ b/command/sub8_missions/missions/align_path_marker.py
@@ -29,23 +29,17 @@ def run(sub):
     s = Searcher(sub, sub.vision_proxies.path_marker.get_pose, pattern)
     resp = None
     print_info.fprint("RUNNING SEARCH PATTERN")
-    resp = yield s.start_search(loop=False, timeout=TIMEOUT_SECONDS)
+    resp = yield s.start_search(loop=False, timeout=TIMEOUT_SECONDS, spotings_req=1)
 
-    if resp is None:
+    if resp is None or not resp.found:
         print_bad.fprint("MARKER NOT FOUND")
         defer.returnValue(None)
     
     print_good.fprint("PATH MARKER POSE FOUND")
-    print_info.fprint("TRANFORMING MARKER POSE TO /map FROM {}".format(resp.pose.header.frame_id))
-    cam_to_baselink = yield sub._tf_listener.get_transform('/base_link', '/'+resp.pose.header.frame_id)
-    cam_to_map = yield sub._tf_listener.get_transform('/map', '/'+resp.pose.header.frame_id)
-    marker_position = cam_to_map.transform_point(rosmsg_to_numpy(resp.pose.pose.position))
-    marker_orientation = cam_to_map.transform_quaternion(rosmsg_to_numpy(resp.pose.pose.orientation))
-    move = sub.move.set_orientation(marker_orientation).zero_roll_and_pitch()
-    position = marker_position.copy()
+    assert(resp.pose.header.frame_id == "/map")
+    move = sub.move.set_orientation(rosmsg_to_numpy(resp.pose.pose.orientation)).zero_roll_and_pitch()
+    position = rosmsg_to_numpy(resp.pose.pose.position)
     position[2] = move._pose.position[2]
-    position[0] = position[0] + cam_to_baselink._p[0]
-    position[1] = position[1] + cam_to_baselink._p[1]
     move = move.set_position(position)
     print_info.fprint("MOVING TO MARKER POSE")
     yield move.go(speed=0.2)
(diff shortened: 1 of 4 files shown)


command/sub8_launch/launch/subsystems/cameras.launch
```diff
@@ -1,4 +1,4 @@
 <launch>
   <include file="$(find sub8_launch)/launch/subsystems/cameras/launch/front_cameras.launch" />
-  <!-- include file="$(find sub8_launch)/launch/subsystems/cameras/launch/down_cameras.launch" / -->
+  <include file="$(find sub8_launch)/launch/subsystems/cameras/launch/down_cameras.launch" />
```
sentree: Make sure the hostname is used to connect to the camera as opposed to the IP address.

kev-the-dev: Its based in the ID camera. I guess it uses that to find it on the network somehow

DSsoto: Did you run the down camera with this?

DSsoto (approved):

perception/sub8_perception/nodes/path_marker_finder.py
```diff
@@ -63,20 +63,36 @@ def __init__(self):
...
+        self.thresh_saturation_low = rospy.get_param("~thresh_satuation_low", 100)
+        self.min_contour_area = rospy.get_param("~min_contour_area", 100)
+        self.length_width_ratio_err = rospy.get_param("~length_width_ratio_err", 0.2)
+        self.approx_polygon_thresh = rospy.get_param("~approx_polygon_thresh", 10)
+        self.shape_match_thresh = rospy.get_param("~shape_match_thresh", 0.4)
+        self.min_found_count = rospy.get_param("~min_found_count", 10)
+        self.timeout_seconds = rospy.get_param("~timeout_seconds", 2.0)
+        camera = rospy.get_param("~marker_camera", "/camera/down/left/image_rect_color")
+
+        self.state_size = 8 # X, Y, Z, THETA, vx, vy, vz, vtheta
```
mattlangford: Is there a reason you're tracking velocity? Your motion model should always define the velocity to be 0 right?

perception/sub8_perception/nodes/path_marker_finder.py
```diff
@@ -173,12 +198,62 @@ def cb_2d(self, req):
...
+        self.found = False
+        self.last3d = None
+        self.filter.errorCovPre = 1. * np.eye(self.state_size, dtype=np.float32)
+        if state is not None:
+            (x, y, z, theta) = state
+            state = np.array([[x], [y], [z], [theta], [0], [0], [0], [0]], dtype=np.float32)
+            self.filter.statePost = state
+
+    def _update_kf(self, (x, y, z, theta)):
+        if self.last_found_time is None:
```
mattlangford: Can you add comments to each of these if statements

perception/sub8_perception/nodes/path_marker_finder.py
```diff
@@ -173,12 +198,62 @@ def cb_2d(self, req):
...
+            if np.linalg.norm(np.array([x, y, z], dtype=np.float32) - self.last3d[0]) > 15:
+                self._clear_filter((x, y, z, theta))
+                rospy.logwarn("Too far apart, resetting")
+                return
+
+        self.found_count += 1
+        self._update_transition_matrix(dt)
+        measurement = 1.* np.array([x, y, z, theta], dtype=np.float32)
+        predict = self.filter.predict()
+        estimated = self.filter.correct(measurement)
```
mattlangford: What happens when you're estimating a theta of pi/2 but you observe -pi/2?

perception/sub8_perception/nodes/path_marker_finder.py
```diff
@@ -63,20 +63,36 @@ def __init__(self):
...
+        self.measurement_size = 4
+
+        self.filter = cv2.KalmanFilter(self.state_size, self.measurement_size)
+        self.filter.transitionMatrix = 1.* np.eye(self.state_size, self.state_size, dtype=np.float32)
+        self.filter.measurementMatrix = 1. * np.array([[1, 0, 0, 0, 0, 0, 0, 0],
+                                                       [0, 1, 0, 0, 0, 0, 0, 0],
+                                                       [0, 0, 1, 0, 0, 0, 0, 0],
+                                                       [0, 0, 0, 1, 0, 0, 0, 0]], dtype=np.float32)
+        self.filter.processNoiseCov = 1e-5 * np.eye(self.state_size, dtype=np.float32)
+        self.filter.measurementNoiseCov = 1e-4 * np.eye(self.measurement_size, dtype=np.float32)
```
mattlangford: Just by guessing, I'd say the measurement noise would be much larger than this - we'd trust our measurements less than we trust our assumption that the thing isn't moving. Experimentation may find that these values work - and with labeled data you could actually compute the variance numbers directly

mattlangford (commented):
Approval pending the response to your comments.
Have you considered tracking something other than x,y,theta (maybe camera frame points, world frame edge points, or I think they best would be just tracking a direction vector pointing where the thing wants you to go with special case handeling observations that are opposite what you'd expect)? Some of these will remove the possible angle weirdness at the limits and could make more sense (you don't really care about where the thing is, just where it's pointing). 

Anyway all this to say - I hope you learned some stuff implementing this!

perception/sub8_perception/nodes/path_marker_finder.py
kev-the-dev: Yeah I suppose it is pretty low. Again, these are numbers similar to those used by other in opencv examples and seemed to be okay. Once we have longer data sets with the down camera I can play with this param.

perception/sub8_perception/nodes/path_marker_finder.py
kev-the-dev: Yeah that did seem to be a problem. When the marker was oriented perfectly upright it behaved weird. I guess I could just have a heuristic to enforce pi/2 if it's abs is roughly pi/2

perception/sub8_perception/nodes/path_marker_finder.py
kev-the-dev: As I'm measuring the pose in the camera frame, it will move over time. Honestly I don't really know what I'm doing but other people online seem to track velocity for computer vision object tracking

perception/sub8_perception/nodes/path_marker_finder.py
mattlangford: So reading the log from the pool day it sounds like you have issues since the whole thing isn't in your FOV at the same time. Maybe you could change from trying to track the whole marker to just trying to track the 4 corner points in the map frame. That would simplify the filtering down to just 4 points that don't move, and would make since with the problems you see. It could be harder to find the 4 corner points however.

perception/sub8_perception/nodes/path_marker_finder.py
DSsoto: You should probably leave a comment next to params you want to validate and optimize against data.

mattlangford: If it were me, I'd keep track of everything in the map frame. That would make prediction steps easy and you would convert from camera frame to world frame before `self.filter.correct`ing.

mattlangford:
Do you have plans on how this perception will change since you can't see the whole thing at once?

kev-the-dev:
I think the solution to not seeing the path marker in the camera frame is a hardware one. I'm going to continue to develop the node assuming the whole thing will be in frame to do pose estimation.

mattlangford:
Sounds good, I think there are still changes to make with the filter, so I'll wait to merge until then

kev-the-dev:
I got rid of camera stuff so this PR is standalone. I'm still working on the filtering and how to store orientation

command/sub8_missions/missions/align_path_marker.py
```diff
@@ -10,45 +11,57 @@
...
     # Wait for vision services, enable perception
-    print_info.fprint("ACTIVATING PERCEPTION SERVICE")
+    print_info("ACTIVATING PERCEPTION SERVICE")
     sub.vision_proxies.path_marker.start()
 
     pattern = []
-    #pattern = [sub.move.right(1), sub.move.forward(1), sub.move.left(1), sub.move.backward(1),
-     #          sub.move.right(2), sub.move.forward(2), sub.move.left(2), sub.move.backward(2)]
+    if DO_PATTERN:
+        pattern = [sub.move.right(SEARCH_RADIUS_METERS), sub.move.forward(SEARCH_RADIUS_METERS),
```
DSsoto: First: Isn't it a bit odd to have two squares of different sizes but they share one corner? I feel like they should share a centroid.

 Second: why not copy line [50](https://github.com/uf-mil/SubjuGator/pull/229/files#diff-0459ee9d56642acd9132f1ae09d0d034R50) above and use it here?

Third: You should probably add a zrp to the beggining of the pattern.

command/sub8_missions/missions/align_path_marker.py
```diff
@@ -10,45 +11,57 @@
...
+    assert(resp.pose.header.frame_id == "/map")
+
+    move = sub.move
+    position = rosmsg_to_numpy(resp.pose.pose.position)
+    position[2] = move._pose.position[2] # Leave Z alone!
+    orientation = rosmsg_to_numpy(resp.pose.pose.orientation)
+
+    move = move.set_position(position).set_orientation(orientation).zero_roll_and_pitch()
+
+    # Ensure SubjuGator continues to face same general direction as before (doesn't go in opposite direction)
```
DSsoto, line 65: Allow the user to specify this. I should be able to align in the way that is closest to the opposite of the current direction.

perception/sub8_perception/nodes/path_marker_finder.py
```diff
@@ -60,152 +66,283 @@ def __init__(self):
...
-    def sortRect(self, rect):
+    def _sort_rect(self, rect):
         '''
         Given a contour of 4 points, returns the same 4 points sorted in a known way.
         Used so that indicies of contour line up to that in model for cv2.solvePnp
-        p[0] = Top left corner of marker
-        p[1] = Top right corner of marker
-        p[2] = Bottom right corner of marker
-        p[3] = Bottom left cornet of marker
+        p[0] = Top left corner of marker         0--1
```
DSsoto, line 153: Like the visuals

perception/sub8_perception/nodes/path_marker_finder.py
```diff
@@ -60,152 +66,283 @@ def __init__(self):
...
+        p[1] = Top right corner of marker        |  |
+        p[2] = Bottom right corner of marker     |  |
+        p[3] = Bottom left cornet of marker      3--2
+
+        The implementation of this is a little bit of magic, and there may be an easier way
```
DSsoto, line 158: I did also did this for the tropedo board. I just took the centroid and then got 4 displacement vectors  from the centroid to the corners. Using atan2 you can get 4 angles from the vector to the positive x axis. Assuming the angles are all in [-pi, pi], it' is as easy as sorting from least to greatest.

perception/sub8_perception/nodes/path_marker_finder.py
```diff
@@ -60,152 +66,283 @@ def __init__(self):
...
+        ps.point = Point(*tvec)
+        vec3 = Vector3Stamped()
+        vec3.vector.x = self.last2d[1][0]
+        vec3.vector.y = self.last2d[1][1]
+        vec3.header.frame_id = self.cam.tfFrame()
+        vec3.header.stamp = self.image_sub.last_image_time
+        map_vec3 = None
+        map_ps = None
+
+        # Transform pose estimate to map framr
```
DSsoto: frame

perception/sub8_perception/nodes/path_marker_finder.py
```diff
@@ -60,152 +66,283 @@ def __init__(self):
...
             cv2.line(self.last_image, (refs[0][0][0], refs[0][0][1]), (refs[2][0][0], refs[2][0][1]), (0, 255, 0)) # Y axis ref
             cv2.line(self.last_image, (refs[0][0][0], refs[0][0][1]), (refs[3][0][0], refs[3][0][1]), (255, 0, 0)) # Z axis ref
+            self._send_debug_marker()
         return True
-    
-    def get_2d_pose(self, r):
+
+    def _get_pose_2D(self, r):
         '''
         Given a sorted 4 sided polygon, stores the centriod and angle
```
DSsoto: centroid

perception/sub8_perception/nodes/path_marker_finder.py
```diff
@@ -37,6 +38,11 @@ class PathMarkerFinder():
     # Model of four corners of path marker, centered around 0 in meters
     LENGTH = 0.6096 # Longer side of path marker in meters
     WIDTH = 0.0762 # Shorter Side
+
+    # Demensions of the prop we made for testing
```
DSsoto, line 45: Still, don't think it would be useful to have the length and width as part of the service request? It would allow us to use arbitrarily sized strips, especially useful for testing.

DSsoto (changes requested):
Make sure to add docstrings to functions that don't have any (unless super trivial).

command/sub8_missions/missions/align_path_marker.py
kev-the-dev: Just realized these moves will all be defined from the same origin, not relative to each other. Will correct

perception/sub8_perception/nodes/path_marker_finder.py
kev-the-dev: that's the best part

perception/sub8_perception/nodes/path_marker_finder.py
kev-the-dev: It would be pretty tough to make it part of the service request, as the pose estimations are done continuously as frames come in not when the pose is requested. It could be a rosparam, but I prefer it hardcoded so I have to be very intensional about changing it (see a diff in a source file).

DSsoto: I see, what you mean. Although you could just add a `/path_marker/set_geometry` service, and make sure you print out what shape you're looking for.

command/sub8_missions/missions/align_path_marker.py
kev-the-dev: Yeah. I imagine we'll have to add some apriori heuristics for picking the direction like Andy said once we see the competition field.

DSsoto: Right, but those heuristics shouldn't be implemented here. I should just be able to specify either of the two alignment options.

DSsoto (approved):

DSsoto:
I'm good with this. @mattlangford any further comments?

DSsoto:
Merging since there seem to be no further comments.

PR #233, opened 2017-05-05
THRUSTERS: fix timing dependent thruster detection

### LINKED ISSUE #195: Single thruster is not detected on first launch
Thruster with NID of 1 is not detected on the first launch of the sub from reboot. Relaunching the sub makes the thruster appear, and is detected on every subsequent launch until another reboot.

DSsoto:
### Thruster Comms
* Increase the serial port timeout from 0.01 s to 0.1 s.
* Parse the response to poll packets to confirm that the response packet
  is actually from the thruster we believe it to be.
* Instead of checking for about 100 thruster ID's on each port, only check the
  ones that have been declared on the thruster layout.
### ThrusterSpinner
* Changed allowable effort range to [-1, 1] so we can spin in both directions
* Added a thruster discovery mode to the thruster spinner utility so that we can
  find out what thrusters are plugged into a port even if they were not declared
  on the thruster_layout.yaml
* Added reporting average packet turnaround time to discovery mode.
**_Proof:_** 
I swapped the connectors for the BR and FR ports on the back of the sub. I ran
the thruster spinner utility and was told that half of the thrusters were missing.
Then I entered discovery mode by pressing 'C' and was able to find the missing
thrusters.
image

Resolves #195

### DIFF
--- a/drivers/sub8_videoray_m5_thruster/sub8_thruster_comm/thruster_comm.py
+++ b/drivers/sub8_videoray_m5_thruster/sub8_thruster_comm/thruster_comm.py
@@ -10,8 +10,7 @@ import rospy
 
 class ThrusterPort(object):
     _baud_rate = 115200
-    _timeout = 0.01  # How long to wait for a serial response
-    _max_motor_id = 100  # Max motor ID for which to verify the existence
+    _timeout = 0.1  # How long to wait for a serial response
 
     def __init__(self, port_info, thruster_definitions):
         '''Communicate on a single port with some thrusters
@@ -37,10 +36,6 @@ class ThrusterPort(object):
             - The thrusters automatically shut down if they do not recieve a
              command for ~2 seconds
 
-            - Potential source of problem for future users:
-                If the sub is ever using more than 100 thrusters, and you find that the Sub isn't finding them,
-                it is because of max_motor_id
-            - This might also trigger timeout errors in the thruster_driver node spin-up
         TODO:
             --> 0x88, 0x8c: Set slew-rate up and down to something tiny
             --> Send messages without getting status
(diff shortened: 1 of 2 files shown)


drivers/sub8_videoray_m5_thruster/sub8_thruster_comm/thruster_comm.py
```diff
@@ -37,10 +36,6 @@ def __init__(self, port_info, thruster_definitions):
             - The thrusters automatically shut down if they do not recieve a
              command for ~2 seconds
 
-            - Potential source of problem for future users:
-                If the sub is ever using more than 100 thrusters, and you find that the Sub isn't finding them,
```
kev-the-dev, line 41: SubjuGator9 better have 100 thrusters.

utils/sub8_diagnostics/scripts/thruster_spinner.py
```diff
@@ -31,7 +31,10 @@ def ports_from_layout(layout):
             for name in thruster_names:
```
kev-the-dev, line 31: Not relevant to the goal of this PR, but in testing with Ken I found it would be helpful to be able to go reverse in thruster_spinner. I think This just means you have to change the limits on the effort command from [0.0, 1.0] to [-1.0, 1.0]

kev-the-dev (approved):
Seems like a better way to do this, and I assume you tested it so I approve. If the thrusters are plugged in wrong, I think it is better for these to show up as down.

utils/sub8_diagnostics/scripts/thruster_spinner.py
DSsoto: You're right, I'll add this.

utils/sub8_diagnostics/scripts/thruster_spinner.py
```diff
@@ -31,7 +31,10 @@ def ports_from_layout(layout):
...
                 port_dict[name] = thruster_port
 
-            msg += "\n\tMotor id's on port: {}".format(thruster_port.motor_ids_on_port)
+            msg += "\n\tMotor id's on port: {}{}{}".format(
```
mattlangford: Assuming you have a printer defined (`p = text_effects.Printer()`), this becomes `msg += str(p.text("Motor id's...").set_bold.blue(np.sort(...)))`
Could be even easier if msg is already a Printer type.

mattlangford (commented):
Consider adding a part to the thruster checker tool that will scan for thrusters and tell you if something's plugged in but not configured (unless this exists and I'm not seeing it)

utils/sub8_diagnostics/scripts/thruster_spinner.py
DSsoto: How would you do it if msg was a printer type? Tbh, honest this is harder to use than it needs to be because it doesn't have any docstrings.

PR #236, opened 2017-05-09
BLUEVIEW: hotfix for launch file typo #OneLineDiff

kev-the-dev:
* one line change to fix blueview launch file, separated from #229 so this can be fixed while I'm still working on path marker


sentree (approved):
Did `git grep` fail you?

PR #242, opened 2017-05-11
Waypoint validity

RustyBamboo:
Stop sub from going where it shouldn't!


RustyBamboo:
screenshot from 2017-05-11 07-01-22

gnc/sub8_controller/nodes/movement_validity.cpp
```diff
@@ -0,0 +1,81 @@
...
+        "is_waypoint_valid", boost::bind(&WaypointChecker::is_waypoint_valid, this, _1, _2));
+  }
+
+  // TODO: What the error was
+  bool is_waypoint_valid(sub8_msgs::WaypointValidity::Request &req, sub8_msgs::WaypointValidity::Response &resp)
+  {
+    // Where the /depth at
+    double wpDepth = req.wp.position.z;
+    resp.valid = true;
+    if (wpDepth > 0)
```
mattlangford: Maybe make it slightly less than 0 - a waypoint that makes us peak out of the water is bad too. Might be worth finding how tall the sub is above the baselink and putting negative that here

gnc/sub8_controller/nodes/movement_validity.cpp
```diff
@@ -0,0 +1,81 @@
...
+  WaypointChecker() : nh_(ros::this_node::getName())
+  {
+    sub_ = nh_.subscribe<nav_msgs::OccupancyGrid>("/ogridgen/ogrid", 1,
+                                                  boost::bind(&WaypointChecker::ogrid_callback, this, _1));
+    validity_service_ = nh_.advertiseService<sub8_msgs::WaypointValidity::Request, sub8_msgs::WaypointValidity::Response>(
+        "is_waypoint_valid", boost::bind(&WaypointChecker::is_waypoint_valid, this, _1, _2));
+  }
+
+  // TODO: What the error was
+  bool is_waypoint_valid(sub8_msgs::WaypointValidity::Request &req, sub8_msgs::WaypointValidity::Response &resp)
```
mattlangford: I think the approach for this should be different. What if this was a library imported into C3 to reject all bad waypoints instead of being required to have a server running

mattlangford (changes requested):
I like the idea a lot, but I think it should be implemented in C3 via some external include rather than as a service like this. Maybe for Sim that doesn't work as well but the same library could be included there and ran with the carrot follower

gnc/sub8_controller/nodes/movement_validity.cpp
RustyBamboo: I think the reason we decided to do this is have a layer of abstraction so that we could also later incorporate other forms of points validations such as stereo. But idk.

@DSsoto Thoughts?

RustyBamboo: I'm changing it. Making it an include I think is better.

legacy/c3_trajectory_generator/include/waypoint_validity.hpp
```diff
@@ -0,0 +1,23 @@
+#include <ros/ros.h>
+#include <geometry_msgs/PoseStamped.h>
+#include <nav_msgs/OccupancyGrid.h>
+
+#include <opencv2/core/core.hpp>
```
DSsoto, line 6: `core.hpp` is already included from `opencv.hpp`. However I like this way better. I'ts good to know what libraries you're depending on and it makes compilation faster than including all of the other stuff opencv.hpp has that you won't use.

legacy/c3_trajectory_generator/include/waypoint_validity.hpp
```diff
@@ -0,0 +1,23 @@
...
+#include "opencv2/opencv.hpp"
```
DSsoto: These functions need docstrings

legacy/c3_trajectory_generator/src/node.cpp
```diff
@@ -30,16 +33,16 @@ subjugator::C3Trajectory::Point Point_from_PoseTwist(const Pose &pose, const Twi
   tf::Matrix3x3(q).getRPY(res.q[3], res.q[4], res.q[5]);
 
   res.qdot.head(3) = vec2vec(tf::Matrix3x3(q) * vec2vec(xyz2vec(twist.linear)));
-  res.qdot.tail(3) = (Eigen::Matrix3d() << 1, sin(res.q[3]) * tan(res.q[4]),
-                      cos(res.q[3]) * tan(res.q[4]), 0, cos(res.q[3]), -sin(res.q[3]), 0,
-                      sin(res.q[3]) / cos(res.q[4]), cos(res.q[3]) / cos(res.q[4])).finished() *
-                     xyz2vec(twist.angular);
+  res.qdot.tail(3) =
```
DSsoto: Remember you can always turn off clang formating for special code blocks like matrix initializations.

legacy/c3_trajectory_generator/include/waypoint_validity.hpp
```diff
@@ -0,0 +1,23 @@
...
+  ros::Subscriber sub_;
+
+  bool check_if_hit(cv::Point center, cv::Size sub_size);
+
+public:
+  WaypointValidity(ros::NodeHandle &nh);
+
+  void ogrid_callback(const nav_msgs::OccupancyGridConstPtr &ogrid_map);
+
+  void is_waypoint_valid(const geometry_msgs::Pose &waypoint, bool &valid, int &error);
```
DSsoto: Returning a non const int error is not a safe way of error reporting. You should return some sort of enum by value.

legacy/c3_trajectory_generator/src/waypoint_validity.cpp
```diff
@@ -0,0 +1,52 @@
...
+  {
+    valid = false;
+    error = 99;
+  }
+}
+
+WaypointValidity::WaypointValidity(ros::NodeHandle &nh)
+{
+  nh_ = &nh;
+  sub_ = nh_->subscribe<nav_msgs::OccupancyGrid>("/ogridgen/ogrid", 1,
```
DSsoto: Easy challenge: make the callback a lambda. It'll be good for you and improves readability.

perception/sub8_pointcloud/include/Classification.hpp
```diff
@@ -12,16 +12,36 @@
...
+#include <opencv2/core/core.hpp>
+#include "opencv2/opencv.hpp"
+
 #include <visualization_msgs/Marker.h>
 #include <visualization_msgs/MarkerArray.h>
 #include <geometry_msgs/Point.h>
 
+#include <list>
+
 class Classification
```
DSsoto: Add docstrings for these methods

perception/sub8_pointcloud/include/Classification.hpp
```diff
@@ -12,16 +12,36 @@
...
+#include <tf2_sensor_msgs/tf2_sensor_msgs.h>
+
+#include <opencv2/core/core.hpp>
+#include "opencv2/opencv.hpp"
+
 #include <visualization_msgs/Marker.h>
 #include <visualization_msgs/MarkerArray.h>
 #include <geometry_msgs/Point.h>
 
+#include <list>
```
DSsoto: Why are you using a list and not a vector?

perception/sub8_pointcloud/src/Classification.cpp
```diff
@@ -19,4 +19,57 @@ pcl::PointCloud<pcl::PointXYZI>::Ptr Classification::filtered(pcl::PointCloud<pc
...
+{
+  // Sub's position relative to the ogrid
+  cv::Point2d where_sub = cv::Point2d(transform.getOrigin().x() / resolution + mat_ogrid.cols / 2,
+                                      transform.getOrigin().y() / resolution + mat_ogrid.rows / 2);
+
+  // List of intersections
+  std::list<cv::Point> intersections;
+  intersections.push_back(where_sub);
+
+  // Find first hits in a circulering ray
```
DSsoto: circulering ray?

perception/sub8_pointcloud/src/Classification.cpp
```diff
@@ -19,4 +19,57 @@ pcl::PointCloud<pcl::PointXYZI>::Ptr Classification::filtered(pcl::PointCloud<pc
...
+  for (float d_theta = 0.f; d_theta <= 6.28; d_theta += 0.05)
```
DSsoto: Use `mil_tools:PI` (`include <mil_tools/mil_tools.hpp>`). Alternatively there are definitions in cmath or boost. Also for float increments, _to me_, it makes a lot more sense to use a while loop.

perception/sub8_pointcloud/src/OGridGen.cpp
```diff
@@ -85,6 +87,11 @@ void OGridGen::publish_ogrid(const ros::TimerEvent &)
     }
   }
 
+  classification_.zonify(mat_ogrid_, resolution_, transform_);
+
```
DSsoto, line 91: Dead code

legacy/c3_trajectory_generator/src/waypoint_validity.cpp
```diff
@@ -0,0 +1,52 @@
...
+  }
+  return false;
+}
+
+void WaypointValidity::ogrid_callback(const nav_msgs::OccupancyGridConstPtr &ogrid_map)
+{
+  this->ogrid_map_ = ogrid_map;
+}
+
+void WaypointValidity::is_waypoint_valid(const geometry_msgs::Pose &waypoint, bool &valid, int &error)
```
DSsoto: Just noting here that we agreed you would make this function return a 
```c++
std::pair<bool, DanielsOGridEnumClass>
```

DSsoto (changes requested):
Requesting some changes, docstrings, and a readme showing how to use it and what parameters it takes. (A wiki page explaining how this works would be nice but it's not urgent.)

perception/sub8_pointcloud/include/OGridRegionType.hpp
```diff
@@ -0,0 +1,7 @@
+#pragma once
+enum OGRD_ERROR_TYPE
```
DSsoto: Forgot the I in OGRID.

legacy/c3_trajectory_generator/README.md
```diff
@@ -0,0 +1,9 @@
+## C3_Trajectory_Generator
+
+## Waypoint Validity
+The following are the error codes that waypoint validity returns:
+- OCCUPIED = 99
+- UNKNOWN = 50
+- UNOCCUPIED = 0
+- TOO_HIGH = 1
```
DSsoto: I think this should be `ABOVE_SURFACE`. I didn't understand what this was for until I looked at the  `WaypointValidity::is_waypoint_valid` implementation.

legacy/c3_trajectory_generator/include/waypoint_validity.hpp
RustyBamboo: Fixed

legacy/c3_trajectory_generator/src/waypoint_validity.cpp
```diff
@@ -0,0 +1,61 @@
...
+
+  if (waypoint.position.z > 0.2)
+  {
+    return std::make_pair(false, ABOVE_WATER);
+  }
+
+  if(!this->ogrid_map_)
+  {
+    ROS_ERROR("WaypointValidity - Did not recieve any ogrid");
+    return std::make_pair(true, NO_OGRID);
```
DSsoto: What's the rationale for calling this a safe waypoint?

legacy/c3_trajectory_generator/src/waypoint_validity.cpp
```diff
@@ -0,0 +1,61 @@
...
+
+  cv::Point where_sub = cv::Point(waypoint.position.x / ogrid_map_->info.resolution + ogrid_map_->info.width / 2,
+                                  waypoint.position.y / ogrid_map_->info.resolution + ogrid_map_->info.height / 2);
+
+  if (ogrid_map_->data.at(where_sub.x + where_sub.y * ogrid_map_->info.width) == 50)
+  {
+    return std::make_pair(false, UNKNOWN);
+  }
+
+  if (check_if_hit(where_sub, cv::Size(5, 5)))
```
DSsoto: How did you choose the size of the sub? Shouldn't you have a separate bit of code that calculates the size passed to this function that doesn't change if we change the resolution of the grid?

legacy/c3_trajectory_generator/src/waypoint_validity.cpp
```diff
@@ -0,0 +1,61 @@
...
+  if(!this->ogrid_map_)
+  {
+    ROS_ERROR("WaypointValidity - Did not recieve any ogrid");
+    return std::make_pair(true, NO_OGRID);
+  }
+
+  cv::Point where_sub = cv::Point(waypoint.position.x / ogrid_map_->info.resolution + ogrid_map_->info.width / 2,
+                                  waypoint.position.y / ogrid_map_->info.resolution + ogrid_map_->info.height / 2);
+
+  if (ogrid_map_->data.at(where_sub.x + where_sub.y * ogrid_map_->info.width) == 50)
```
DSsoto: Cast the enum value into an int here for comparison.

legacy/c3_trajectory_generator/src/waypoint_validity.cpp
RustyBamboo: yeah I was thinking about this for a while. I think the only case in which we wouldn't have an ogrid is if we don't have the blueview -- but we probably would still want to test/compete and move around.

But yeah, idk, it might be more reasonable to disable movement if we don't have an ogird.

legacy/c3_trajectory_generator/src/node.cpp
```diff
@@ -191,12 +235,34 @@ struct Node {
...
 
+    // Check if we will hit something while in trajectory the new trajectory
+    geometry_msgs::Pose traj_point;  // Convert messages to correct type
+    auto p = c3trajectory->getCurrentPoint();
+    traj_point.position = vec2xyz<Point>(p.q.head(3));
+    quaternionTFToMsg(tf::createQuaternionFromRPY(p.q[3], p.q[4], p.q[5]), traj_point.orientation);
+
+    std::pair<bool, WAYPOINT_ERROR_TYPE> checkWPResult =
+        waypoint_validity_.is_waypoint_valid(Pose_from_Waypoint(current_waypoint));
+    if (checkWPResult.first == false && checkWPResult.second == WAYPOINT_ERROR_TYPE::OCCUPIED)
```
DSsoto: Is this the only error type that we should be concerned about?

legacy/c3_trajectory_generator/src/waypoint_validity.cpp
```diff
@@ -0,0 +1,61 @@
+#include "waypoint_validity.hpp"
+
+// Point and sub_size must be relative to ogrid, IE: meters * 1/OGRID_RESOLUTION + offset
```
DSsoto: The units here are not clear. Ogrid resolution is a tuple: `(rows x cols)`. What does it mean to have a unit of `meters` * `unitless ordered pair`?

Maybe you meant that it should be in terms of the size of a single grid in the ogrid.

legacy/c3_trajectory_generator/src/waypoint_validity.cpp
```diff
@@ -0,0 +1,61 @@
...
+bool WaypointValidity::check_if_hit(cv::Point center, cv::Size sub_size)
+{
+  for (int x = center.x - sub_size.width / 2; x < center.x + sub_size.width / 2; ++x)
+  {
+    for (int y = center.y - sub_size.height / 2; y < center.y + sub_size.height / 2; ++y)
+    {
+      if (ogrid_map_->data.at(x + y * ogrid_map_->info.width) == 99)
```
DSsoto: Cast the value of the enum into an int here for the comparison. You don't want this to break if you change the enum.

legacy/c3_trajectory_generator/src/node.cpp
```diff
@@ -167,21 +181,51 @@ struct Node {
...
-          !goal->uncoordinated);
+          Point_from_PoseTwist(goal->posetwist.pose, goal->posetwist.twist), goal->speed, !goal->uncoordinated);
       current_waypoint_t = now;
       this->linear_tolerance = goal->linear_tolerance;
       this->angular_tolerance = goal->angular_tolerance;
+
+      // Check if waypoint is valid
+      std::pair<bool, WAYPOINT_ERROR_TYPE> checkWPResult =
+          waypoint_validity_.is_waypoint_valid(Pose_from_Waypoint(current_waypoint));
+      if (checkWPResult.first == false)  // got a point that we should not move to
```
DSsoto: It doesn't seem like you've handle all possibilities that will take this branch through the code. For example, waypoints above the water are not handled.

legacy/c3_trajectory_generator/src/node.cpp
RustyBamboo: I think so. I can't think of a case where the trajectory will go above the water height -- because this would only happen if the waypoint was set to go above the water, which is blocked. And I can't think of a case where the trajectory will suddenly enter an unknown region. If we were able to set a waypoint, then it is unoccupied, and only occupied grids can appear, which would stop the trajectory. And I think if we block unknown trajectories, we may have scenarios where rotating in place will not be allowed.

perception/sub8_pointcloud/include/OGridRegionType.hpp
```diff
@@ -0,0 +1,7 @@
+#pragma once
```
DSsoto: Is this used at all?

I think it's best to not have multiple copies of this and just use [the one in C3](https://github.com/uf-mil/SubjuGator/pull/242/files#diff-732da72f004e537bc7a7d42305096915R7)

legacy/c3_trajectory_generator/src/waypoint_validity.cpp
RustyBamboo: http://docs.ros.org/jade/api/nav_msgs/html/msg/MapMetaData.html

The map resolution is m/pixel

legacy/c3_trajectory_generator/src/node.cpp
RustyBamboo: Yeah, I was not sure what the correct height cap to set, and so I decided not to check for it.

legacy/c3_trajectory_generator/src/node.cpp
DSsoto: I see what you mean, although I'm not sure we want to proceed quietly when we are navigating without occupation information.

DSsoto:
Can you add a parameter to C3 to disable OGRID checking?

Sorry the review process is taking a while, but for something that can stop the sub from moving, this is how it needs to be.

legacy/c3_trajectory_generator/src/waypoint_validity.cpp
DSsoto: For this very reason I think we should be able to turn off all ogrid checks from C3 if we need to. But then it would be very explicit what we're doing.

perception/sub8_pointcloud/include/OGridRegionType.hpp
RustyBamboo: I think it would be better to separate the too. In one, we are using the enum purely to draw an ogrid. In the c3, the enum is used to represent error types, that are not all necessarily tied into the ogrid. Also, I ran into a struggle of including headers from different packages.

legacy/c3_trajectory_generator/src/waypoint_validity.cpp
DSsoto: Then it would be clearer to say that the sub_size is in units of map cells. (meters cancel out)

legacy/c3_trajectory_generator/src/node.cpp
DSsoto: You can set it to the water surface for now, and then we'll see if we want to change it to something different after testing.

You always want to make sure that your'e handling all of the possibilities when you start branching off with `if` statements.

legacy/c3_trajectory_generator/src/waypoint_validity.cpp
RustyBamboo: Good idea

perception/sub8_pointcloud/include/OGridRegionType.hpp
DSsoto: I can help you out with including headers from different packages, lol don't be intimidated by cmake.

We should talk about this design more in person. I see your point, but it seems like these enums will be encoding the same information.

mattlangford:
Should it be a ros param or a flag in the waypoint action request to disable validity checks?

RustyBamboo:
I was thinking it would be better to leave like a layer so that waypoint validation could also be used in the sim

mattlangford:
A flag would still work for sim right? On Navigator we had the -f flag and the `force` keyword for the action request to not check feasibility.
Also consider a case where you want to do two consecutive moves where you want to check validity for the first but not the second, it seems like having to call a service between the moves isn't the right way to accomplish that.

RustyBamboo:
I don't think I understand what the problem is...

So if we wanted to check validity for one waypoint (and the trajectory there), we would set the rosparam to true. Once the action client receives a success message, it should set the param to false and give another waypoint, which would then be a blind move and ignore validity check for both waypoint and trajectory.

mattlangford:
Yeah that seems like a weird way to do it. What if the something cancels during the invalid rosparam time and the next move is issued with no checks? You'd have to explicitly set the param before each movement to enforce valid movements only (incase someone or something set the ignore param and didn't clear it) which sounds like something that C3 should just take as an argument for each move.

RustyBamboo:
Ah... Okay.
Well, I was thinking this would be more of a thing which we disable when we don't have blueview mounted.

mattlangford:
Wouldn't that be up to the waypoint validator - always return true if there's no valid ogrid?

legacy/c3_trajectory_generator/src/waypoint_validity.cpp
DSsoto: :(

DSsoto (approved):

DSsoto:
@RustyBamboo I'm happy with this now (assuming you're using the right size for the sub).

@mattlangford care to update your review?

legacy/c3_trajectory_generator/src/node.cpp
```diff
@@ -191,12 +242,35 @@ struct Node {
       c3trajectory_t = now;
     }
 
-    while (c3trajectory_t + traj_dt < now) {
-      c3trajectory->update(traj_dt.toSec(), current_waypoint,
-                           (c3trajectory_t - current_waypoint_t).toSec());
+    // Remember the previous trajectory
+    auto old_trajectory = c3trajectory->getCurrentPoint();
```
mattlangford: Can you explain this? Does it just let us stop if our plan encounters a new obstacle? If so, I think you should rename it since it sounds like this will take us to the previous waypoint.

legacy/c3_trajectory_generator/src/node.cpp
```diff
@@ -167,22 +181,59 @@ struct Node {
...
+          ROS_ERROR("can't move there! - waypoint is above water");
+          current_waypoint = old_waypoint;
+        }
+      }
     }
-    if (actionserver.isPreemptRequested()) {
+    if (actionserver.isPreemptRequested())
+    {
       current_waypoint = c3trajectory->getCurrentPoint();
+      current_waypoint.do_waypoint_validation = c3trajectory->do_waypoint_validation;
```
mattlangford: You should probably always set the current waypoint to never have waypoint validation if a movement was canceled - though I don't think it breaks anything to have it.

legacy/c3_trajectory_generator/src/node.cpp
```diff
@@ -167,22 +181,59 @@ struct Node {
...
+        {
+          ROS_ERROR("can't move there! - waypoint is occupied");
+          current_waypoint = old_waypoint;
+        }
+        // if point is above water, reject move
+        if (checkWPResult.second == WAYPOINT_ERROR_TYPE::ABOVE_WATER)
+        {
+          ROS_ERROR("can't move there! - waypoint is above water");
+          current_waypoint = old_waypoint;
+        }
```
mattlangford, line 248: Do you think we should cancel right away and return unsuccessfully here with an error code to whomever requested this action?

legacy/c3_trajectory_generator/src/node.cpp
```diff
@@ -191,12 +242,35 @@ struct Node {
...
       c3trajectory_t += traj_dt;
     }
 
+    // Check if we will hit something while in trajectory the new trajectory
+    geometry_msgs::Pose traj_point;  // Convert messages to correct type
+    auto p = c3trajectory->getCurrentPoint();
+    traj_point.position = vec2xyz<Point>(p.q.head(3));
+    quaternionTFToMsg(tf::createQuaternionFromRPY(p.q[3], p.q[4], p.q[5]), traj_point.orientation);
+
+    std::pair<bool, WAYPOINT_ERROR_TYPE> checkWPResult =
```
mattlangford, line 278: This only really checks the location we are currently at (plus some small forward offset) right, or does this check the end goal? If it's the former, is it needed?

perception/sub8_pointcloud/include/Classification.hpp
```diff
@@ -12,16 +12,37 @@
...
   Classification(ros::NodeHandle *nh);
 
+  // Usage: statistical filterting to get rid of outliers and noise
   pcl::PointCloud<pcl::PointXYZI>::Ptr filtered(pcl::PointCloud<pcl::PointXYZI>::ConstPtr pointCloud);
+
+  // Usage: Find all the first occupied points in an expanding circle, then color the ogrid
+  void zonify(cv::Mat &mat_ogrid, float resolution, const tf::StampedTransform &transform);
+
+  // Usage: Draw a fake ogrid to test in SIM
+  void fake_ogrid(cv::Mat &mat_ogrid, float resolution, const tf::StampedTransform &transform);
```
mattlangford, line 47: This seems kind of strange to me, shouldn't something external be generating ogrids even for sim?

legacy/c3_trajectory_generator/src/waypoint_validity.cpp
```diff
@@ -0,0 +1,64 @@
...
+  int sub_x = 1.5 / ogrid_map_->info.resolution;
+  int sub_y = 1.5 / ogrid_map_->info.resolution;
+  if (check_if_hit(where_sub, cv::Size(sub_x, sub_y)))
+  {
+    return std::make_pair(false, WAYPOINT_ERROR_TYPE::OCCUPIED);
+  }
+
+  if (ogrid_map_->data.at(where_sub.x + where_sub.y * ogrid_map_->info.width) == (uchar)WAYPOINT_ERROR_TYPE::UNKNOWN)
+  {
+    return std::make_pair(false, WAYPOINT_ERROR_TYPE::UNKNOWN);
```
mattlangford: I think you should allow movements to unknown areas (for example a point behind the sub) until we are sure something is there

legacy/c3_trajectory_generator/src/waypoint_validity.cpp
```diff
@@ -0,0 +1,64 @@
...
+void WaypointValidity::ogrid_callback(const nav_msgs::OccupancyGridConstPtr &ogrid_map)
+{
+  this->ogrid_map_ = ogrid_map;
+}
+
+// Convert waypoint to be relative to ogrid, then do a series of checks (unknown, occupied, or above water).
+// Returns a bool that represents if the move is safe, and an error
+std::pair<bool, WAYPOINT_ERROR_TYPE> WaypointValidity::is_waypoint_valid(const geometry_msgs::Pose &waypoint, bool do_waypoint_validation)
+{
+  if(!do_waypoint_validation) return std::make_pair(true, WAYPOINT_ERROR_TYPE::UNKNOWN);
```
mattlangford: Maybe have a code for waypoints that weren't checked?

legacy/c3_trajectory_generator/src/waypoint_validity.cpp
```diff
@@ -0,0 +1,64 @@
...
+
+  // Area we want to check around the sub
+  int sub_x = 1.5 / ogrid_map_->info.resolution;
+  int sub_y = 1.5 / ogrid_map_->info.resolution;
+  if (check_if_hit(where_sub, cv::Size(sub_x, sub_y)))
+  {
+    return std::make_pair(false, WAYPOINT_ERROR_TYPE::OCCUPIED);
+  }
+
+  if (ogrid_map_->data.at(where_sub.x + where_sub.y * ogrid_map_->info.width) == (uchar)WAYPOINT_ERROR_TYPE::UNKNOWN)
```
mattlangford: Will this throw an error if the point is outside the ogrid?

mattlangford (commented):
Some questions.
Also it might be nice to edit the sub singleton pose proxy to allow blind movements.

legacy/c3_trajectory_generator/src/waypoint_validity.cpp
DSsoto: I think in the end we'll be checking against the persistent occupancy grid so this behavior should be ok.

legacy/c3_trajectory_generator/src/node.cpp
RustyBamboo: Yeah, I'll add an error code in the MoveTo action

legacy/c3_trajectory_generator/src/node.cpp
RustyBamboo: True

legacy/c3_trajectory_generator/src/node.cpp
RustyBamboo: This checks collision during actual movement to the defined current_waypoint

legacy/c3_trajectory_generator/src/waypoint_validity.cpp
RustyBamboo: Added

legacy/c3_trajectory_generator/src/node.cpp
mattlangford: But only a small delta forward right?

legacy/c3_trajectory_generator/src/waypoint_validity.cpp
RustyBamboo: Yeah, I added a try-catch and made it return an occupied error

perception/sub8_pointcloud/include/Classification.hpp
RustyBamboo: lol. nothing in SIM, so i've been used fake_ogrid for testing

legacy/c3_trajectory_generator/src/node.cpp
RustyBamboo: Yeah, so each loop the sub sets a new waypoint a small delta forward and I check if it's occupied or not, and if it is, I don't apply the small delta

RustyBamboo:
@mattlangford what do mean by allowing blind movements in pose proxy?

sub.move.go(blind=True)?

RustyBamboo:
https://stackoverflow.com/questions/28828957/enum-to-string-in-modern-c-and-future-c17-c20
Fun stuff.

legacy/c3_trajectory_generator/src/node.cpp
mattlangford: The small delta is like centimeters forward though isn't it?

mattlangford:
Yeah, but it looks like everything is set up for that so nevermind. Just to confirm, the blind flag defaults to false is nothing is specified?

legacy/c3_trajectory_generator/src/node.cpp
mattlangford: Regardless, I think that a check for the validity of the goal at each timer event is something that should be checked, right?

RustyBamboo:
Yes it will be false, because of ROS defaults to 0

legacy/c3_trajectory_generator/src/node.cpp
RustyBamboo: Right now if the waypoint becomes occupied, it will go as close as it can to the point and then throw a TRAJECTORY_OCCUPIED. Originally, I had it where it did check if the goal become occupied each loop but I'm not sure we would want to stop just because of noise or a diver, etc.

mattlangford: But in the nominal case where the sub is properly tracking, the goal we are following should be inside the sub, like it's just a cm or two infront if where the previous waypoint was. That means that by the time the C3 carrot hits something we would have already hit it.

RustyBamboo: The waypoint validity checks for any occupied grids in a 1.5m * 1.5m area

mattlangford: Yeah you're right, that would give it some buffer room. I'd still be worried that the sub going full speed can't stop that quickly if somehow the initial validity check passed. I'll defer to @DSsoto as to whether this is fine.

mattlangford (approved):

legacy/c3_trajectory_generator/src/node.cpp
DSsoto: 1.5 m is about twice the length of the sub. I might even be too large actually. I'm fine with this for now, we'll get feedback from testing.

PR #243, opened 2017-05-13
CAMERAS: add calibration for down camera, enable rectification

kev-the-dev:


DSsoto (approved):
This is not a great calibration, but until we get that screw to lock down the focus on the lens, this will do.

Issue #238, opened 2017-05-11
Increase window size for bus-voltage kill


DSsoto:
Right now, the bus_voltage kill triggers unnecesarily due to the amount of fluctuation in the bus voltage readings.

On our small pool outing on May 10, I noticed that when station-keeping in still waters, the bus voltage readings had a total spread of about 0.5 volts in a span of just about 2 seconds.

Although I did not record numbers, I would venture to say that the average spread during move commands and other transient thruster behaviour is about 2 volts.

I propose that we only kill the sub when the average bus voltage drops below the pre-defined limits for a window of about 30 seconds.

I will submit a PR doing this soon, but open to hearing other ideas/counterarguments.

mattlangford:
Kalman filter bus voltage!

DSsoto:
There isn't one bus voltage, all of the bus voltage packets from different thrusters spam to the same /bus_voltage topic. I doubt we can assume measurements with Gaussian error of an average bus voltage in this scenario. **_Ad-hoc!!_**      (**_tldr_**: overkill)

DSsoto:
Addressed by #247. The window size is currently 30 seconds but can be adjusted based on observations from the next pool day.

PR #247, opened 2017-05-13
BUS VOLTAGE: widen window for bus voltage estimate

DSsoto:
* Implement a BusVoltageEstimator class which for now just takes an average
  of the readings from the last n seconds from all thrusters together
* Refactor code related to bus voltage to use the new BusVoltageEstimator
* Moved around functions and renamed variables, functions to imporve readability /
  keep functions with logically close functionality together
* Update sub8_videoray_m5_thruster readme
* Addresses #238

### DIFF
--- a/drivers/sub8_videoray_m5_thruster/nodes/thruster_driver.py
+++ b/drivers/sub8_videoray_m5_thruster/nodes/thruster_driver.py
@@ -14,9 +14,58 @@ from sub8_thruster_comm import thruster_comm_factory
 from ros_alarms import AlarmBroadcaster, AlarmListener
 lock = threading.Lock()
 
+class BusVoltageEstimator(object):
+    '''
+    Class that estimates sub8's thruster bus voltage.
+    As of May 2017, this is just a simple rolling average with a constant width sliding
+    window. However add_reading and get_estimate methods are left for when smarter
+    filtering is needed
+    '''
+    class VoltageReading(object):
+        def __init__(self, voltage, time):
+            self.v = voltage
+            self.t = time
+
+    def __init__(self, window_duration):
+         '''
+         window_duration - float (amount of seconds for which to keep a reading in the buffer)
+         '''
+         self.window_duration = rospy.Duration(window_duration)
+         self.last_update_time = None
+         self.buffer = []
+
+    def add_reading(self, voltage, time):
+        ''' Adds voltage readings to buffer '''
+        self.buffer.append(self.VoltageReading(voltage, time))
+        self.last_update_time = time
+        self.prune_buffer()
+
+    def prune_buffer(self):
+        ''' Removes readings older than the window_duration from buffer '''
+        for reading in self.buffer:
+            age = rospy.Time.now() - reading.t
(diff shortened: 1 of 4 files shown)


drivers/sub8_videoray_m5_thruster/nodes/thruster_driver.py
```diff
@@ -14,9 +14,58 @@
...
+    def get_last_update_time(self):
+        ''' Returns time of most recent reading in buffer '''
+        if self.last_update_time is None:
+            self.last_update_time = rospy.Time.now()
+        return self.last_update_time
 
 class ThrusterDriver(object):
-    _dropped_timeout = 1 # s
+    _dropped_timeout = 1.0 # s
+    _window_duration = 30.0 # s
```
kev-the-dev, line 68: I feel like 30 seconds is kind of big. We still do want it to kill quickly if the battery drops. Something like 10s seems more reasonable.

drivers/sub8_videoray_m5_thruster/readme.md
```diff
@@ -4,16 +4,17 @@ VideoRay M5 Thruster Driver
 # Authors
 
 This code was originally authored by Forrest Voight (forrestv) and Matthew Griessler.
+It is now maintained by David Soto.
 
 # Purpose
 
-This node handles the interface between the sub's control computer and the thrusters. It communicates via serial to each thruster, individually issuing commands.
+This node handles the interface between the sub's control computer and the thrusters. It communicates via serial to each thruster, individually issuing commands, it also receives response packets from each of th thrusters (including current, voltage, rpm, faults, etc.) and publishes them.
```
kev-the-dev: > th thrusters

kev-the-dev (approved):
The window seems a little big to me. We're really still looking to just throw out an outlier reading right? We still want to kill pretty quickly if the battery is actually low.

drivers/sub8_videoray_m5_thruster/nodes/thruster_driver.py
DSsoto: I disagree. I can see it killing after a 10 seconds where we would have come right back above the kill voltage at the end of some move.

However, running it for 20 extra seconds after we actually get to a killable (steady state) voltage does not make much of a difference.

DSsoto:
The bus voltage actually drops momentarily during transient periods with high thrust output, but then recovers after it. It's more like something to filter transient periods than a single reading outlier remover.

sentree:
The way I understand our batteries (which could be completely wrong) is that there is a cutoff voltage where we should stop using them. Even if they only dip below that voltage while moving, isn't that still bad for the batteries in the long term?

DSsoto:
We can ask @duggerd what the optimal value is. I think as it is right now, there is a safety margin above that value where it is harmful for the batteries. I've had the sub kill at about 45.5 volts because it momentarily dipped below 44 volts (kill voltage).

DSsoto:
Merging since there have been no further comments, and we can always just modify the kill voltage.

PR #248, opened 2017-05-13
GNC: update hard-soft magnetic calibration

DSsoto:
Calibration data collected at Towne Parc pool on May 13. Down camera box and Imaging Sonar was mounted. (No passive sonar hardware or pneumatics)
> original error: 0.9788%
> error: 0.5374%


kev-the-dev (approved):
Looks good.

Issue #245, opened 2017-05-13
Chain Sequence of Pose Commands


LucasBA:
Mechanism for giving the sub a list of pose commands to achieve from the command line and from the mission runner.

mattlangford:
C3 will accept velocity only waypoints, this can be used to go in one direction without slowing down at the end. You'd just need to track the subs position and switch the waypoint at some point to get it to move where you want.

DSsoto:
This is dealt with in #249

PR #249, opened 2017-05-14
MISSIONS: add class to execute consecutive moves

LucasBA:
#245 An example of how to use it is pose_sequence_test.py

It points the sub at the waypoint drives there and sets the new orientation

### DIFF
--- a/command/sub8_missions/missions/pose_sequence_test.py
+++ b/command/sub8_missions/missions/pose_sequence_test.py
@@ -1,15 +1,24 @@
 from txros import util
 import numpy as np
+from mil_misc_tools import text_effects
 from sub8.sub_singleton import PoseSequenceCommander
 
+fprint = text_effects.FprintFactory(title="SEQUENCE_TEST").fprint
 
 @util.cancellableInlineCallbacks
 def run(sub_singleton):
-    print "Starting"
-    poses = [np.array([-1, 0, 0, 0, 0, 3]),
-    		np.array([-1, 0, 0, 0, 0, 0]),
-    		np.array([-1, 0, 0, 0, 0, 3]),]
-    poses_command = PoseSequenceCommander(sub_singleton,poses)
-    yield poses_command.go_to_sequence()
-    
-    print "Done!"
+    '''Used for testing in sim. These commands may not be safe for the real sub.'''
+    fprint("Starting Test With Eulers", msg_color="blue")
+    positions = [np.array([-1, 0, 0]), np.array([0, 0, 3]),
+    	np.array([-1, 0, 0]), np.array([0, 0, 3])]
+    orientations = [np.array([-1, 0, 0]), np.array([0, 0, 3]),
+    	np.array([-1, 0, 0]), np.array([0, 0, 3])]
+    poses_command = PoseSequenceCommander(sub_singleton)
+    yield poses_command.go_to_sequence_eulers(positions, orientations)
+    fprint("Done With Eulers!", msg_color="blue")
+
+    fprint("Starting Test With Quaternions", msg_color="blue")
+    orientations = [np.array([0, 0, 0, 1]), np.array([0, 0, 0, 1]),
+	np.array([0, 0, 0, 1]), np.array([0, 0, 0, 1])]
+    yield poses_command.go_to_sequence_quaternions(positions, orientations)
+    fprint("Done with Quaternions", msg_color="blue")
--- a/command/sub8_missions/missions/pose_sequence_test.py
+++ b/command/sub8_missions/missions/pose_sequence_test.py
@@ -1,0 +1,15 @@
+from txros import util
+import numpy as np
+from sub8.sub_singleton import PoseSequenceCommander
+
+
+@util.cancellableInlineCallbacks
+def run(sub_singleton):
+    print "Starting"
+    poses = [np.array([-1, 0, 0, 0, 0, 3]),
+    		np.array([-1, 0, 0, 0, 0, 0]),
+    		np.array([-1, 0, 0, 0, 0, 3]),]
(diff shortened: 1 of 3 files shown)


command/sub8_missions/sub8/sub_singleton.py
```diff
@@ -312,6 +312,17 @@ def _run_look(self, spotings_req):
 
             yield self.sub.nh.sleep(.5)
 
+class PoseSequenceCommander(object):
+    def __init__(self, sub, pose_array):
+        self.sub = sub
+        self.pose_array = pose_array
+
+    @util.cancellableInlineCallbacks
+    def go_to_sequence(self):
```
RustyBamboo: Add a comment on how to use it

command/sub8_missions/sub8/sub_singleton.py
```diff
@@ -312,6 +312,17 @@ def _run_look(self, spotings_req):
 
             yield self.sub.nh.sleep(.5)
 
+class PoseSequenceCommander(object):
```
RustyBamboo, line 315: Would be cool to also have command line passing

RustyBamboo (commented):
Looks good. I'd add a quick a comment on how to use it. Would it be possible to add a speed argument?

DSsoto:
Can you squash the last two commits into the first one? 
I imagine you just forgot about them but those are not acceptable commit messages.

LucasBA:
Yeah I meant to use - - amend and I think I missed a -

command/sub8_missions/sub8/sub_singleton.py
LucasBA: Will do

command/sub8_missions/sub8/sub_singleton.py
LucasBA: I was hoping to add that in the next pr, I it has some usefulness without it

command/sub8_missions/missions/pose_sequence_test.py
```diff
@@ -0,0 +1,15 @@
+from txros import util
+import numpy as np
+from sub8.sub_singleton import PoseSequenceCommander
+
+
+@util.cancellableInlineCallbacks
+def run(sub_singleton):
+    print "Starting"
```
DSsoto: Can you use a mil_misc_tools.text_effects printer? (Examples [here](https://github.com/uf-mil/SubjuGator/search?utf8=%E2%9C%93&q=fprint&type=) and [here](https://github.com/uf-mil/SubjuGator/search?utf8=%E2%9C%93&q=printer&type=))
This is the colorful printer we like to use for high level mission stuff and it's good for you to start getting used to it.

command/sub8_missions/missions/pose_sequence_test.py
```diff
@@ -0,0 +1,15 @@
...
+    poses = [np.array([-1, 0, 0, 0, 0, 3]),
+    		np.array([-1, 0, 0, 0, 0, 0]),
+    		np.array([-1, 0, 0, 0, 0, 3]),]
+    poses_command = PoseSequenceCommander(sub_singleton,poses)
```
DSsoto: We tend to represent poses as 3-vectors concatenated with quaternions. Can you make it look like this?:
```python
PoseSequenceCommander(sub_singleton,positions,orientations)
```
 Alternatively you could concatenate them but you 
should raise an Exception with a helpful error msg if it is not of length 7.

command/sub8_missions/sub8/sub_singleton.py
```diff
@@ -312,6 +312,17 @@ def _run_look(self, spotings_req):
...
+    def __init__(self, sub, pose_array):
+        self.sub = sub
+        self.pose_array = pose_array
+
+    @util.cancellableInlineCallbacks
+    def go_to_sequence(self):
+        for pose in self.pose_array:
+            yield self.sub.move.look_at_without_pitching(np.array(pose[0:3])).go()
+            yield self.sub.move.relative(np.array(pose[0:3])).go()
+            yield self.sub.move.set_orientation(transformations.quaternion_multiply(self.sub.pose.orientation,transformations.quaternion_from_euler(pose[3], pose[4], pose[5]))).go()
```
DSsoto: Keep in mind, python code will soon stop passing build checks if it is not PEP8. You might want to run this through flake8.

DSsoto (changes requested):
Requesting a few changes.

LucasBA:
I think I made all the changes.

command/sub8_missions/sub8/sub_singleton.py
LucasBA: Are you sure we want command line passing? It seems like the list of poses on one line will get pretty unwieldy after 2 or 3. It might be a lot easier to just edit a python script.

DSsoto: I think that's the use case we're concerned with, 2 or 3. It would be great to have.

LucasBA: Do we want to use the same base input format as move_command?

DSsoto: Yeah, It would be nice to be able to run something like this from the terminal:
```shell
subc yl 90deg f 1 u 0.5
# or also the long versions
```

LucasBA: In that case I think the best way to go about it is to put the current move_command in a for loop

LucasBA: I just added that

kev-the-dev:
I left some comments on these files where they were duplicated in #257. Please refer to those.

DSsoto (approved):

DSsoto:
This addresses #245

PR #250, opened 2017-05-14
SCRIPTS: switch bag variable to image_rect_color, add fps alias

kev-the-dev:
We found with Navigator that it's more convientent to have image_rect_color in bags since all perception nodes (should) use it instead of image_raw. This makes it easier to test without having to run image_proc, and you could pottentialy still unrectify back to image_raw if needed. So this switches the bag variables to use image_rect_color.

I also added the alias I was using at testing today to see if the cameras were up. It just runs rostopic hz on the camera topics so we can know which ones are up and what their framerates are. Useful for diagnostics.

### DIFF
--- a/scripts/bagging_variables.sh
+++ b/scripts/bagging_variables.sh
@@ -1,9 +1,9 @@
 #!/bin/bash
 
-# This script will be sourced by the bmux command and the variables will be
-# used by the bag command. It is imperative that each variable is in the bag_*
-# namespace and contains a space diliniated list of ROS topics. The only two
-# exceptions to this rule are the BAG_ALWAYS and BAG_DIR variables.
+# This script will be sourced by the bag command as a configuration file. It is
+# imperative that each variable is in the bag_* namespace and contains a space
+# deliniated list of ROS topics. The only two exceptions to this rule are the
+# BAG_ALWAYS and BAG_DIR variables.
 
 
 # Define topics that should be in every bag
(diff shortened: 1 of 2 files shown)


scripts/bagging_variables.sh
```diff
@@ -12,12 +12,14 @@ export BAG_ALWAYS="/odom /absodom /clock /tf /tf_static"
 # Define the directory that the bags will be stored in
 export BAG_DIR=~/bags
 
+# Which camera image type to use
+export CAM_SUFFIX="image_rect_color"
 
 # Topic variables that can be used from the bag command
-export bag_front_left_cam="/camera/front/left/camera_info /camera/front/left/image_raw"
-export bag_front_right_cam="/camera/front/right/camera_info /camera/front/right/image_raw"
+export bag_front_left_cam="/camera/front/left/camera_info /camera/front/left/$CAM_SUFFIX"
```
DSsoto: Can you add a cli option to `bag` to set `CAM_SUFFIX`? I'm fine with having image rect color as the default.

DSsoto (approved):
This is fine, but I'd prefer if you added my request.

scripts/bagging_variables.sh
whispercoros: If you have a look at `scripts/bagging_variables.sh`, you should see some variables in all capital letters at the top of the file. Are you suggesting to make the camera suffix one of these?

scripts/bagging_variables.sh
```diff
@@ -12,12 +12,14 @@ export BAG_ALWAYS="/odom /absodom /clock /tf /tf_static"
 # Define the directory that the bags will be stored in
 export BAG_DIR=~/bags
 
+# Which camera image type to use
+export CAM_SUFFIX="image_rect_color"
```
sentree: Please note that a command line argument should be added to `mil_common/scripts/bag.sh` for the `CAM_SUFFIX` variable. This can be copied from the sections that address the `BAG_ALWAYS` or `BAG_HOME` variables. Also, I would prefer that it be called `BAG_CAM_SUFFIX` to keep with my naming convention in this file as well as to have the comment begin with the word "Define" (sorry to nit pick).

sentree (changes requested):
I would appreciate it if you could copy the comment header for this file from `mil_common/scripts/bag.sh` as it has been changed.

scripts/bagging_variables.sh
kev-the-dev: Good idea, but that would be done in mil_common and probably best by @whispercoros

scripts/bagging_variables.sh
kev-the-dev: If you want to add that flag  for CAM_SUFFIX (or, better, arbitrary meta variables) it would be cool. Until than it was just a way to save me typing the same thing thrice and we shouldn't over think it. I'll change the variable name and comment.

sentree: Alright, I will add this to the list and merge this once it is done.

sentree (approved):
All of my comments were addressed.

PR #251, opened 2017-05-15
VIZ: move robot description param setting

DSsoto:
* Move the setting of the robot_description parameter from the shore_control
  launch to both the sub8 launch and the gazebo launch


sentree (approved):
This looks like it gets the job done, but perhaps it would be best to have an interface launch file that both the sub and simulation launch files use. Of course this only makes sense if there are multiple things that are shared, but there probably are.

kev-the-dev (approved):
Good

PR #261, opened 2017-05-22
GAZEBO: add start gate, nav gate, octagon

Grymessa:
screenshot from 2017-05-22 12-34-41
screenshot from 2017-05-22 12-35-02
screenshot from 2017-05-22 12-35-38
Octagon, Nav gate, Start gate in order of appearance. As a side note, I am still a bit concerned about sizing. Gazebo doesn't have a way of telling me the size in meters or any useful unit of measurement and I had to scale down the original models significantly. I used the size of the sub as a comparison for most of these when it came to scale. The painting is all correct but the octagon has some odd shading stuff going on with it. Might be a result of the scaling. I was hoping I could fix it with blender but doesn't look like it. I am downloading solid works now to see what I can't do since I have the files. With that in mind, it might be worth waiting to merge this? I'll leave that call to you guys. I'm not super confident in my CADing skills. The colors are correct and I think the sizes are in the ballpark just not exact. 

As a side note, apparently the torpedo board has changed for this season so for the model, as far as we could read it, it needs to be in the shape of a tentacle and squid? Not sure but that's the main reason I excluded it from this update. Additionally, I need a stand for the pingers I think so those can be retrieved and "buried". Currently working on that now.


sentree:
If you click the object inside of Gazebo, it will give you a side bar where you can get the exact size in meters of the mesh. Make sure to double check that things are accurately sized since we will be using this to test perception.

Grymessa:
@whispercoros As far as I can tell, that method only works if the properties are defined in the file. Which you would think would be the case with the auto generated DAE files but apparently not. I believe it is supposed to be located under link->collision->geometry but it isn't. If I import a pre-generated model from Gazebo however, it is there. So something must be missing when I export from Blender. 

And this all became a moot point because I figured out how to measure things in Blender. Turns out I'm just mildly incompetent. -updating models now. Luckily textures scale. 

EDIT: Sizes should be fixed now. That just leaves the tentacle boards to be made and then we have all the mission modules in. That and the stand for the little treasure bits that sits underneath the octogon I believe. I will start work on the gravity for the bits now.

simulation/sub8_gazebo/README.md
```diff
@@ -18,6 +18,10 @@ And these can be combined if you'd like.
...
 
 ### To add custom textures to models
+This guide assumes you have created a mesh and texture within gazebo. All the models I have been given from Solidworks have already come with meshes so I did not have to create them. Look up guides on how to bind textures to UV Maps in Blender. That is what you will be doing for the most part to create your texture files. You'll want to save the texture file as a separate png for this. An interesting side note about Blender, the way the UV Maps work and meshes work, you cannot scale an object differently in each dimension and preserve the mesh. The entire object must be scaled at the same ratio (.5,.5,.5 for example). Otherwise when you export the object you will encounter many terrible errors trying to load it into Gazebo. 
+
+tldr; Scale your objects evenly in blender or else the machine spirits will be displeased.
```
sentree, line 23: They're everywhere!

simulation/sub8_gazebo/launch/duck.launch
```diff
@@ -40,6 +40,27 @@
     -x -22.3 -y 14.16 -z -4.47 -Z -1.57 -model lid"
     output="screen" />
 
+  <!-- Attempting to spawn the start gate -->
+  <node
+      name ="start_gate" pkg="gazebo_ros" type="spawn_model"
+      args="-file $(find sub8_gazebo)/models/start_gate/start_gate.sdf -sdf
```
sentree, line 46: I think in the near future, we may want to consider removing the old `sub8_simulation` package and renaming `sub8_gazebo` to `sub8_simulation`. Just an idea for now. Thoughts, @DSsoto?

sentree (approved):
Everything looks good here, this should really help us with testing while we are waiting on a stable thruster configuration.

mattlangford:
We had the same problem with the weird shadings on the sub model at one point (might still happen). I'd guess the problem is that normals aren't flipped the right way (easy to verify) or maybe gazebo wants faces to be triangles instead of rectangular (I think there's an export option for this).

A ghetto method for measuring inside of gazebo (which I've never proved to be correct) is spawning a cube - I think it spawns a 1m^3 cube

simulation/sub8_gazebo/launch/duck.launch
DSsoto: That's not a bad idea. I don't think there's anything in `sub8_simulation` that would be used in the future (but do look into it).

lol Then again, this doesn't actually help anybody.

Issue #253, opened 2017-05-15
Make path_marker_finder generic


DSsoto:
The competition requires finding the pose of multiple bright orange polygons objects repeatedly.

Kevin has implemented this successfully for the path marker (#229) but there is no reason to not use it for the bin lids as well.

I propose changing the name of the perception code from `path_marker_finder` to `orange_rectangle_finder` or some other more generic name. It should have a /set_geometry service that allows you to set the width and height of the object you want to find the pose of.

sentree:
I can look in to this if @ironmig doesn't want to since I will be handling this part of the perception for the bins challenge anyway.

kev-the-dev:
I got this. Already doing that so it can work with unit tests

PR #263, opened 2017-05-22
PATH MARKER: add service to set model geometry, use mil_common RectFinder

kev-the-dev:
* replace relevant part of path_marker_finder with RectFinder from uf-mil/mil_common#52
* use solvePnP rotation estimate instead of 2D
* add alias to launch rviz with sub's config file
* add service to path_marker_finder to set model geometry so it can be adhoc used with to scale models of the path_marker and in the future other orange rectangular object like the bins lid
* rename path_marker to orange_rectangle_finder as per #253

### DIFF
--- a/command/sub8_launch/config/orange_rectangle_finder.yaml
+++ b/command/sub8_launch/config/orange_rectangle_finder.yaml
@@ -1,17 +1,36 @@
+# True to provide debug image and rviz Markers for debugging
 debug_ros: True
+
+# paramaters to pass to canny contour detection.
+#   runs: cv2.Canny(img, canny_low, canny_low*self.canny_ratio)
 canny_low: 100
 canny_ratio: 3.0
+
+# Threshold hue and saturation values to create mask to find orange objects.
+#   runs: cv2.inRange(img, (0, thresh_saturation_low, 0), (thresh_hue_high, 255, 255))
 thresh_hue_high: 60
 thresh_satuation_low: 100
+
+# Minimum number of pixels in area of a contour to not be filtered
 min_contour_area: 100
+
+# multiple of contour arcLength to pass to OpenCV approxPolygon to find 4 sided polygon
 epsilon_factor: 0.02
+
+# Maximum value from cv2.matchshapes to be considered a valid contour
 shape_match_thresh: 0.4
+
+# <min_found_count> number of images where object is found within <timeout_seconds> to be considered found
 min_found_count: 10
 timeout_seconds: 2.0
+
+# True to provide 3d pose approximation service using PnP
 do_3D: True
-marker_camera: "/camera/down/left/image_rect_color"
+
+# Topic to pull images from to find rectangle
+image_topic: "/camera/down/left/image_rect_color"
 
 # initialize to correct scale model of path marker
 # missions should call /vision/path_marker/set_geometry with correct model
-length: 1.2192
-width: 0.1524
+length: 1.2192 # longer side of rectangle in meters
+width: 0.1524  # shorter side of rectangle in meters
--- a/command/sub8_launch/config/orange_rectangle_finder.yaml
+++ b/command/sub8_launch/config/orange_rectangle_finder.yaml
@@ -8,7 +8,10 @@ epsilon_factor: 0.02
 shape_match_thresh: 0.4
 min_found_count: 10
 timeout_seconds: 2.0
-length: 1.2192
-width: 0.1524
 do_3D: True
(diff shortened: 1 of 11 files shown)


command/sub8_launch/config/orange_rectangle_finder.yaml
```diff
@@ -0,0 +1,17 @@
+debug_ros: True
```
mattlangford, line 2: Most of this is self explanatory, but comments briefly saying what each does would be good. Also some of them could use unit definitions too

command/sub8_launch/config/orange_rectangle_finder.yaml
```diff
@@ -0,0 +1,17 @@
...
+canny_ratio: 3.0
+thresh_hue_high: 60
+thresh_satuation_low: 100
+min_contour_area: 100
+epsilon_factor: 0.02
+shape_match_thresh: 0.4
+min_found_count: 10
+timeout_seconds: 2.0
+do_3D: True
+marker_camera: "/camera/down/left/image_rect_color"
```
mattlangford: More like `image_topic` right?

command/sub8_launch/launch/subsystems/perception.launch
```diff
@@ -1,16 +1,21 @@
...
+  <group ns="vision">
+    <node if="$(arg buoy)" pkg="sub8_perception" type="buoy_2d.py" name="buoys" output="screen">
+      <rosparam file="$(find sub8_launch)/config/buoy_2d.yaml" command="load" ns="buoys" />
+    </node>
 
-  <node pkg="sub8_perception" type="path_marker_finder.py" name="path_marker_finder" output="screen">
-    <rosparam file="$(find sub8_launch)/config/path_marker_finder.yaml" command="load" />
-  </node>
+    <node if="$(arg orange_rectangle)" pkg="sub8_perception" type="orange_rectangle_finder.py" name="orange_rectangle">
+      <rosparam file="$(find sub8_launch)/config/orange_rectangle_finder.yaml" command="load" />
```
mattlangford, line 17: Do the rules call it orange rectangle? I thought they called it path marker or navigation marker or sonething

command/sub8_missions/missions/align_path_marker.py
```diff
@@ -18,6 +19,14 @@
 SPEED = 0.5
 MISSION="Align Path Marker"
 
+# Model of four corners of path marker, centered around 0 in meters
+LENGTH = 1.2192 # Longer side of path marker in meters
```
mattlangford, line 23: Maybe pull these from the param?

mattlangford: Scratch that, I see what you're doing below

perception/sub8_perception/nodes/orange_rectangle_finder.py
```diff
@@ -279,30 +230,30 @@ def _update_kf(self, (x, y, z, dy, dx)):
...
-        i_points = np.array(p,dtype=np.float)
-        # Use camera intrinsics and knowledge of marker's real demensions to get a pose estimate in camera frame
-        _, rvec, tvec =  cv2.solvePnP(PathMarkerFinder.PATH_MARKER, i_points, self.cam.intrinsicMatrix(), np.zeros((5,1)))
+    def _get_pose_3D(self, corners):
+        tvec, rvec = self.rect_model.get_pose_3D(corners, cam=self.cam, rectified=True)
         if tvec[2][0] < 0.3: # Sanity check on position estimate
             rospy.logwarn("Marker too close, must be wrong...")
             return False
+        rmat, _ = cv2.Rodrigues(rvec)
+        vec = rmat.dot(np.array([1, 0 , 0]))
```
mattlangford: You can also take the first column of the rmat to get the forward vector (which is what this dot product does)

mattlangford (changes requested):
No big problems. Just want to verify the rules call it an orange rectangle not a path marker?

command/sub8_launch/launch/subsystems/perception.launch
kev-the-dev: One objective of this PR is to address #253, to make this perception node generic so it can be used to find the two challenges involving orange rectangles, the path marker and the lid of the bins challenge.

kev-the-dev:
Comments addressed!

mattlangford:
Cool, you could make it color agnostic as well if you wanted too.

mattlangford (approved):

PR #265, opened 2017-05-22
TF REPUB: wait for tf and catch timeout error

DSsoto:


command/sub8_launch/scripts/tf_republisher.py
```diff
@@ -13,7 +14,11 @@ def got_range(msg):
     if rospy.Time.now() < rospy.Time(0.5):
         listener.clear()
     t = rospy.Time(0)
-    trans, rot = listener.lookupTransform("/base_link", "/map", t)
+    try:
+        listener.waitForTransform('/base_link', '/map', t, rospy.Duration(1))
+        trans, rot = listener.lookupTransform("/base_link", "/map", t)
+    except (tf.Exception, tf.LookupException, tf.ConnectivityException, tf.ExtrapolationException) as e:
```
kev-the-dev, line 20: you only really need ```tf.Exception``` because the rest are subclasses but whatevas

kev-the-dev (approved):
Good. We should always swallow TF errors

command/sub8_launch/scripts/tf_republisher.py
DSsoto: Oh shit, you're right. I mindlessly copy-and-pasted that whole parenthesis. :disappointed:

PR #266, opened 2017-05-23
VideoRay Propellers

DSsoto:
* This adds the effort to thrust mapping for the VideoRay m5 thrusters running their original propeller prototype (as opposed to @jnez71's).
* It also makes the B-matrix vectors line up with the standard defined in the [sub map](https://drive.google.com/file/d/0B2qRA9f_CN_9TndLZUNseURVUEk/view).
* Changes sim thrust bounds to match new props used IRL.
* Adds odom to the kill bag topics
* Version controls firmware settings

_**Note:**_ these changes were tested on the sub. We are able to station keep with these changes but we intability while moving because one of the horizontal thrusters (6-FLH) is outputting much less thrust than commanded.

TODO: make it possible for different thrusters to use different effort-to-thrust mappings by listing the calibration file individally for each thruster in the thruster_layout. This would allow us to easily use mixed prop design configurations in case that ever becomes a necessity.


drivers/sub8_videoray_m5_thruster/launch/thruster_driver.launch
```diff
@@ -1,4 +1,4 @@
 <launch>
     <node pkg="sub8_videoray_m5_thruster" type="thruster_driver.py" name="thruster_driver"
-        args="--calibration_path=$(find sub8_videoray_m5_thruster)/config/calibration.json" />
+        args="--calibration_path=$(find sub8_videoray_m5_thruster)/config/json_nez_prop.json" />
```
kev-the-dev: Aren't we running with the stock props now? So shouldn't the config be for that?

DSsoto: I guess, this doesn't matter as of now because we need to address issue #267 before being able to run the sub in the current configuration (both stock and Jason's).

command/sub8_alarm/alarm_handlers/kill.py
```diff
@@ -22,8 +22,8 @@ def cleared(self, alarm):
     def bagger_dump(self):
         """Call online_bagger/dump service"""
         camera = '/camera/front/left/camera_info /camera/front/left/image_raw '
-        navigation = '/wrench /wrench_actual /c3_trajectory_generator/trajectory_v \
-                      /c3_trajectory_generator/waypoint /trajectory '
+        navigation = '/wrench /wrench_actual /c3_trajectory_generator/trajectory_v' + \
+                     '/c3_trajectory_generator/waypoint /trajectory /odom'
```
kev-the-dev, line 26: That's kosher, but it would be better if these topics were pulled from the ```BAG_ALWAYS``` and ```bag_*``` variables if available so we didn't have two copies of magic topics.

drivers/sub8_videoray_m5_thruster/config/firmware_settings/2.txt
```diff
@@ -0,0 +1,41 @@
+
```
kev-the-dev, line 1: For diff-ibility, these firmware files should have the same whitespace before and after the firmware settings

drivers/sub8_videoray_m5_thruster/config/firmware_settings/readme.md
```diff
@@ -0,0 +1,7 @@
+# VideoRay Thruster Firmware Configuration
+This directory contains the firmware configurations meant to be used with our VideoRay m5 thrusters.
```
kev-the-dev, line 2: It would be nice if this README has instructions on how to update these files if firmware is suspected to have changed

kev-the-dev (approved):

drivers/sub8_videoray_m5_thruster/config/firmware_settings/readme.md
DSsoto: You mean how to change the firmware back? Because you can update these files in any plain ole' editor.

kev-the-dev: I meant explain the sequence of commands/actions needed to enter config mode and print out this data for each thruster.

DSsoto: That's documented in docs.ufl.mil

kev-the-dev:
Fix the format issue failing the build and we can merge

command/sub8_alarm/alarm_handlers/kill.py
sentree: I agree with this, it makes sense to have a central location for the bagging variables, even if that means creating a `bag_kill` variable and making it a combination of the ones we already have.

drivers/sub8_videoray_m5_thruster/config/firmware_settings/readme.md
sentree: If the documentation is external, it should probably be linked or referred to in the readme.

drivers/sub8_videoray_m5_thruster/nodes/thruster_driver.py
```diff
@@ -183,6 +183,10 @@ def publish_bus_voltage(self, *args):
 
     def check_bus_voltage(self, voltage):
         ''' Raises bus_voltage alarm with a corresponding severity given a bus voltage '''
+        # Timing bug occurs here occasionally so I (David) wil disable enforcing bus_voltage kills
```
sentree, line 186: No need for attribution here. I'm sure we will all be running `git blame` when problems arise, don't you worry... Also, there is a spelling error on this line.

drivers/sub8_videoray_m5_thruster/nodes/thruster_driver.py
```diff
@@ -263,7 +267,7 @@ def command_thruster(self, name, force):
             Example names are BLR, FLH, etc.
         '''
         target_port = self.port_dict[name]
-        margin_factor = 0.8
+        margin_factor = 1.0  # Not sure why this would not be anything but one - David + Jason
```
sentree, line 270: If we have agreed that this should be 1.0, then we should probably just remove the variable altogether.

drivers/sub8_videoray_m5_thruster/readme.md
```diff
@@ -11,10 +11,11 @@ It is now maintained by David Soto.
 This node handles the interface between the sub's control computer and the thrusters. It communicates via serial to each thruster, individually issuing commands, it also receives response packets from each of the thrusters (including current, voltage, rpm, faults, etc.) and publishes them.
 
 # Calibration
-The thrusters take an input on [-1, 1]. The calibration.json file provided maps commanded effort values to produced output thrust in Newtons.
+The thrusters take an input on [-1, 1]. The files provided  in `config/calibration/` map commanded effort values to expected output thrust in Newtons.
```
sentree, line 14: If @jnez71's propellers ever make a comeback, can we please name the calibration file `jason.json`?

command/sub8_alarm/alarm_handlers/kill.py
mattlangford: Side note, it looks like you're missing a space. AFAIK `+` doesn't add a space

DSsoto:
Closing this PR in favor of #290.

PR #268, opened 2017-05-23
FORMATTING: clang-format c/c++ files in the repo

DSsoto:
I made sure that everything compiled and was able to move around without issues in gazebo.

I'd like to run stuff on the sub before I'm comfortable merging this.


kev-the-dev (approved):
Got same output when running clang on my computer. Should be good. clang-format can't really break anything that wouldn't be caught at compile time.

sentree (approved):
These changes all worked in the simulator.

PR #271, opened 2017-05-26
WAYPOINT: remove error spam when no ogrid present

RustyBamboo:


DSsoto:
Please don't close PR's without saying why. Also, it would be nice to get a throttled mesage only when the sub is executing a non-blind move and the grid is missing.

PR #273, opened 2017-05-29
THRUSTER MAPPER: fix the slsqp solver configuration

jnez71:
* Make regularization weight as small as possible to avoid leaving B*u=w plane
* Decrease convergence tolerance to avoid numerical error
* Initialize solver to a clipped version of the analytical unconstrained solution
* Explain this in the docstrings a bit


DSsoto (approved):
Looks good, and @jnez71 showed me this publishing good wrenches to /wrench_actual in the sim.

Issue #272, opened 2017-05-26
Add imaging sonar to simulation and integrate ogrid


kev-the-dev:
After merging #242, we noticed we'll need a simulated imaging sonar to test out navigation stack now that it checks for way point validity.

PR #274, opened 2017-06-01
SIM: Add lidarscan to simulate blueview

RustyBamboo:
Reference to #272 
- remove no ogrid spam
- add gazebo interface to convert lidarscan to BlueViewScan
- rescale the ogrid to fit the pool

### DIFF
--- a/legacy/c3_trajectory_generator/src/waypoint_validity.cpp
+++ b/legacy/c3_trajectory_generator/src/waypoint_validity.cpp
@@ -42,7 +42,6 @@ std::pair<bool, WAYPOINT_ERROR_TYPE> WaypointValidity::is_waypoint_valid(const g
 
   if (!this->ogrid_map_)
   {
-    ROS_ERROR("WaypointValidity - Did not recieve any ogrid");
     return std::make_pair(false, WAYPOINT_ERROR_TYPE::NO_OGRID);
   }
 
--- a/perception/sub8_pointcloud/launch/ogrid.launch
+++ b/perception/sub8_pointcloud/launch/ogrid.launch
@@ -1,8 +1,13 @@
 <launch>
     <node pkg="sub8_pointcloud" type="ogrid_generator" name="ogrid_pointcloud" >
         <rosparam>
+            # meters per "pixel"
             resolution: 0.2
+            
+            # width and height of ogrid in meters 
             ogrid_size: 150
+
+            # Ignore blueview imaging sonar points below depth in meters
             pool_depth: 3.0
 
             # How many points should be allowed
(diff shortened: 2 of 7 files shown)


RustyBamboo:
screenshot from 2017-06-01 14-37-32

perception/sub8_pointcloud/launch/ogrid.launch
```diff
@@ -2,7 +2,7 @@
     <node pkg="sub8_pointcloud" type="ogrid_generator" name="ogrid_pointcloud" >
         <rosparam>
             resolution: 0.2
-            ogrid_size: 50.0
+            ogrid_size: 150
```
kev-the-dev: Parameters like this should have a comment describing what they are and what unit's they're in.

legacy/c3_trajectory_generator/src/waypoint_validity.cpp
```diff
@@ -42,7 +42,6 @@ std::pair<bool, WAYPOINT_ERROR_TYPE> WaypointValidity::is_waypoint_valid(const g
 
   if (!this->ogrid_map_)
   {
-    ROS_ERROR("WaypointValidity - Did not recieve any ogrid");
```
kev-the-dev, line 45: Ahhh but I liked the red death on my terminal :(

simulation/sub8_gazebo/nodes/gazebo_controller.py
```diff
@@ -66,6 +71,16 @@ def publish_height(self, msg):
             )
         )
 
+    def publish_blueview_ping(self, msg):
+        self.blueview_ranges_pub.publish(
+            BlueViewPing(
+                header=msg.header,
+                bearings=[(msg.angle_min + i * msg.angle_increment) for i in xrange(len(msg.ranges))],
+                ranges=msg.ranges,
+                intensities=[500 for x in msg.ranges]
```
kev-the-dev: LaserScan messages have intensity values. Why aren't we using those?

kev-the-dev (changes requested):
Exciting! Going to be nice having the ability to test imaging sonar code in ideal conditions for bug fixing.

simulation/sub8_gazebo/nodes/gazebo_controller.py
RustyBamboo: For some reason the intensities are all 0... I've been trying to figure out why that is, but I assume it's not part of the ray sensor. I could maybe make the intensities the ranges? Thoughts?

kev-the-dev: Wait why the ranges? Just so it visualizes better? Are the intensities used anywhere in the ogrid stack?

RustyBamboo: When I was setting the blueview intensities to be the laser intestines, the pointcloud wouldn't show up in rviz because all the intensities would be 0. Setting it to ranges is better than just a constant value I think, it it makes it a rainbow! The intensities aren't being used anywhere in the ogrid stack.

kev-the-dev (approved):

PR #275, opened 2017-06-01
BUOY: refactor and fix mission and perception (backup plan)

kev-the-dev:
* refactor much of the buoy finder python node. I found many bugs and added documentation. I also added support for the LAB color space, and set new thresholds in LAB based on the bags I could find. Though it still just uses color to segment, it seems to work pretty well now is various lightning conditions, but a better approach would be easy to drop in.
* rewrite the Buoy mission. It makes better use of txros now and should solve the mission faster in some cases (for example, it asynchronously stops searching as soon as it has a pose on a buoy, then continues searching after it is bumped)
* Fix a problem in sim where odom start at [0, 0, 0], meaning that some underwater moves will be canceled because they are positive z. Odom in sim now starts at [0, 0, initial z]. 
* some other small bugs in vision tools and launch files have been fixed


command/sub8_missions/missions/buoy.py
```diff
@@ -2,132 +2,98 @@
...
+        self.sub.vision_proxies.buoy.start()
+
+        pattern = self.pattern()
+        self.do_search = True
+        search = self.search()
+        while len(self.bumped) != len(self.found):
+            for color in self.found:
+                if self.found[color] is not None and color not in self.bumped:
+                    pattern.cancel()
+                    yield self.bump(color)
```
kev-the-dev: So right now it just bumps buoys as it finds them. This will actually not maximize points as order matters. I'll work on this.

kev-the-dev: Did it!

perception/sub8_perception/nodes/buoy_finder.py
```diff
@@ -0,0 +1,364 @@
...
+            self.color_space = color_space
+            low = '/color/buoy/{}/{}_low'.format(self.color, color_space)
+            high = '/color/buoy/{}/{}_high'.format(self.color, color_space)
+            if not rospy.has_param(low):
+                continue
+            self.thresholds = [np.array(rospy.get_param(low)),
+                               np.array(rospy.get_param(high))]
+            rospy.loginfo("BUOY - Thresholds for {} buoy loaded using {} colorspace".format(
+                          self.color, self.color_space))
+            return
```
mattlangford: Should this `return` be here? The debug stuff won't work unless there are no color space params.

perception/sub8_perception/nodes/buoy_finder.py
```diff
@@ -0,0 +1,364 @@
...
+                self.thresholds[i][i2] = float(val)
+                rospy.loginfo("SETTING {} thresholds[{}][{}]={}".format(self.color, i, i2, self.thresholds[i][i2]))
+            for i in range(len(self.thresholds[0])):
+                cv2.createTrackbar('low {}'.format(self.color_space[i]), self.color, int(self.thresholds[0][i]), 255,
+                                   lambda x, _ind=i: change_threshold(0, _ind, x))
+            for i in range(len(self.thresholds[1])):
+                cv2.createTrackbar('high {}'.format(self.color_space[i]), self.color, int(self.thresholds[1][i]), 255,
+                                   lambda x, _ind=i: change_threshold(1, _ind, x))
+
+    def segment(self, img):
```
mattlangford: This isn't really a segment function, more like `get_mask`

perception/sub8_perception/nodes/buoy_finder.py
```diff
@@ -0,0 +1,364 @@
...
+        trans = np.array(t)
+        R = mil_ros_tools.geometry_helpers.quaternion_matrix(rot_q)
+
+        if (buoy.last_t is None) or (np.linalg.norm(trans - buoy.last_t) > 0.1):
+            buoy.last_t = trans
+            buoy.observations.append(true_center)
+            buoy.pose_pairs.append((t, R))
+
+        if len(buoy.observations) > self.min_observations:
+            self.buoys[buoy_type].est = self.multi_obs.lst_sqr_intersection(buoy.observations, buoy.pose_pairs)
```
mattlangford: I still don't think this a proper way to do this. The buoys do move under water. Last year they had a buoy attached to something on the surface and that one moved around like crazy. Even with truncating the length of the observations at 300, you'll keep 20 seconds or so of history. From my experience last year, the buoy could move a decent amount in that time.

mattlangford (commented):
This is the backup plan? What's the primary plan?

perception/sub8_perception/nodes/buoy_finder.py
kev-the-dev: Yeah... maybe expect n observations in x minutes, than clear the buffer if it hasn't been observed in x minuets?

perception/sub8_perception/nodes/buoy_finder.py
kev-the-dev: True.

kev-the-dev:
This is really just getting the old threshold based buoy stuff working again. Ideally we will be using @aaron-iglesias's machine learning solution for segmentation if it performs well.

perception/sub8_perception/sub8_vision_tools/rviz.py
```diff
@@ -14,8 +14,8 @@ class RvizVisualizer(object):
     '''Cute tool for drawing both depth and height-from-bottom in RVIZ
     '''
 
-    def __init__(self):
-        self.rviz_pub = rospy.Publisher("visualization/projection", visualization_msgs.Marker, queue_size=3)
+    def __init__(self, topic="visualization/projection"):
```
DSsoto: Can you use a more specific topic name here?

DSsoto (approved):
Looks good, did you catch anything that needed to be fixed last pool day?

kev-the-dev:
Just pushed the little fixes from test day. Still runs well in sim.

perception/sub8_perception/sub8_vision_tools/rviz.py
kev-the-dev: This is just the default. Individual nodes (like buoy_finder) will override it. I think we should go through and get all perception nodes publishing to the same rviz marker topic later

DSsoto: I disagree about publishing all markers to the same topic, unless everyone publishes to both a specific and a general topic. It's often very useful to reduce clutter as much as possible when visualizing.

This is ok though.

PR #276, opened 2017-06-06
SIM: magic missile

Grymessa:
Added teleport functionality to sim. Coordinates to various mission objects found in README of Gazebo sim.


command/sub8_missions/nodes/move_command
```diff
@@ -39,6 +45,34 @@ def main(args):
...
+                z = args.gteleport[2]
+                modelstate = ModelState(
+                    model_name='sub8',
+                    pose = Pose(position=Point(x,y,z), orientation=Quaternion(0,0,0,0)),
+                    twist = Twist(linear=Point(0,0,0),angular=Point(0,0,0)),
+                    reference_frame= 'world')
+                    #Sometimes you need to sleep in order to get the thing to publish
+                    #Apparently there is a latency when you set a publisher and it needs to actually hook into it.
+                    #Wierd ROS stuff.
+                #print("Before")
```
kev-the-dev: We're trying to get rid of debug / dead code in the repo. If you don't imagine ever having these lines uncommented in "production" code, you should remove them

command/sub8_missions/nodes/move_command
```diff
@@ -39,6 +45,34 @@ def main(args):
...
+                #kill sub
+                ab = AlarmBroadcaster('kill', node_name='kill')
```
kev-the-dev: Because this script uses the txros client api, you should use the [txros version of ROS alarms classes](https://github.com/uf-mil/ros_alarms/blob/master/ros_alarms/tx_alarms.py). Using both rospy thing and txros has caused problems in the past

simulation/sub8_gazebo/README.md
```diff
@@ -18,21 +18,31 @@ And these can be combined if you'd like.
...
-11. Here you shall place bind_material and bind your materials to the mesh, applying your texture to the object. 
-12. Copy the format as shown in colored_lid.dae. 
+11. Here you shall place bind_material and bind your materials to the mesh, applying your texture to the object.
+12. Copy the format as shown in colored_lid.dae.
 
 **NOTE:** I have not tested trying to apply more than one texture. To create the texture, I used Blender. This created a texture wrapped around the mesh. I exported the texture file as a png as you can see in the bin folder of models. The image you use should be in the same directory as the dae file. I have not tested putting it elsewhere. Additionally, if you see yellow and black lines on your object as opposed to your  texture, it means Gazebo/dae file could not find you texture! Check to make sure the name of the image matches the name of the file being imported. If you see a standard grey object, this means you did not import or code the  texture properly. Recheck the steps and compare to either the skybox or the colored_lid dae files.  
+
+### Teleport Pose Coordinates for Challenges
+
+1. Bin Assembly - -24 14 -3.5
```
kev-the-dev: Good to have these here for now, but eventually we should have some file (YAML/ XML) which defines these positions (and orientations) in a way a script can parse. Then we should have a nice interface to ```subc -gt buoy``` or something similar. Also, if this is your eventual plan it is nice to leave a ```TODO:``` comment somewhere to describe this.

simulation/sub8_gazebo/src/sub8_state_set.cpp
```diff
@@ -82,6 +88,6 @@ void StatePlugin::Init()
 /////////////////////////////////////////////////
 void StatePlugin::OnUpdate()
 {
-  auto fixed_pose = last_ref_pose_ + first_pose_;
-  model_->SetWorldPose(fixed_pose);
+  //TODO ADD CHECK, LISTEN FOR AN UPDATE ON SET_MODEL_STATE
```
kev-the-dev: This TODO can be removed

simulation/sub8_gazebo/src/sub8_state_set.cpp
```diff
@@ -3,6 +3,9 @@
 #include "gazebo/common/Assert.hh"
 #include "gazebo/common/Events.hh"
 #include "ros/ros.h"
+#include "gazebo/msgs/MessageTypes.hh"
```
kev-the-dev: Doesn't look like you need either of these last two imports.

simulation/sub8_gazebo/worlds/a_whole_new.world
```diff
@@ -17,7 +17,7 @@
       <uri>model://sun</uri>
     </include>
 
-    <!-- Never delete this -->
+    <!-- Never delete this, uncomment to bring back the ghost cessna
```
kev-the-dev, line 20: Ahhh I didn't catch your loophole last PR. It says don't delete, but it never said anything about commenting out :)

command/sub8_missions/nodes/move_command
```diff
@@ -95,6 +129,8 @@ if __name__ == '__main__':
         help="Make end goal have zero roll and pitch.")
     parser.add_argument('-b', '--blind', action='store_true',
         help="Do not check if waypoint is safe.")
+    parser.add_argument('-gt', '--gteleport', type=float,nargs='+',
+        help="Teleport sub in Gazebo to coordinates. Does nothing if not in Gazebo.")
```
kev-the-dev: you can add ```metavar='X Y Z'``` here to make it more clear from the help prompt how to enter the coordinates.

command/sub8_missions/nodes/move_command
```diff
@@ -39,6 +45,34 @@ def main(args):
         fprint("Odom found!", msg_color='green')
 
         action_kwargs = {'speed': float(args.speed), 'blind': bool(args.blind)}
+        if args.gteleport:
+            try:
```
kev-the-dev: You should check here that 3 arguments were given, and print something descriptive if they weren't so you don't just get a mysterious out of bounds exception.

kev-the-dev (changes requested):
Cool! This is going to make SIM much more useful for mission testing (the days of making coffee while I wait for SubjuGator to move across the simulated transdec are over).  Found a couple of things to note in coding practices and convention

simulation/sub8_gazebo/README.md
Grymessa: I need to figure out how to do these scripts, never written one before. This is likely my next "project".

simulation/sub8_gazebo/worlds/a_whole_new.world
Grymestone: Just amended the pull request, that should fix everything save the YAML scripts which I am now attempting to implement.

Grymessa:
Just amended the pull request, that should fix everything save the YAML scripts which I am now attempting to implement.

Grymessa:
Yaml now working, new squid based torpedo boards in, alongside Elp table.

command/sub8_missions/nodes/move_command
```diff
@@ -37,8 +47,49 @@ def main(args):
         fprint("Waiting for odom...")
         yield sub.tx_pose()
         fprint("Odom found!", msg_color='green')
-
+        rospack = rospkg.RosPack()
+        config_file = os.path.join(rospack.get_path('sub8_missions'), 'nodes', 'tp.yaml')
+        f = yaml.load(open(config_file, 'r'))
         action_kwargs = {'speed': float(args.speed), 'blind': bool(args.blind)}
+        if args.gteleport:
```
kev-the-dev: I couldn't get this to run. If only ```subc -gt "buoy"``` is entered, the code can never get to this line. How do you use it?

command/sub8_missions/nodes/move_command
```diff
@@ -37,8 +47,49 @@ def main(args):
...
+            try:
+                state_set_pub = yield nh.advertise('gazebo/set_model_state', ModelState)
+                try:
```
kev-the-dev: Generally you don't use try except for stuff like this. Better to check if it is valid manually

```
split_args = args.gteleport.split()
if len(split_args) == 3
    # run as coordinates
else:
    # check if in yaml
```

command/sub8_missions/nodes/move_command
```diff
@@ -37,8 +47,49 @@ def main(args):
...
+                    x = float(args.gteleport.split(' ')[0])
+                    y = float(args.gteleport.split(' ')[1])
+                    z = float(args.gteleport.split(' ')[2])
+                except Exception:
+                    fprint("Format should be 'X Y Z' if you wish to input coords. Checking directory for saved location", msg_color='yellow')
+                    try:
+                        argz = f[args.gteleport]
+                        print(argz)
```
kev-the-dev: debug printline

command/sub8_missions/nodes/move_command
```diff
@@ -37,8 +47,49 @@ def main(args):
...
         action_kwargs = {'speed': float(args.speed), 'blind': bool(args.blind)}
+        if args.gteleport:
+            try:
+                state_set_pub = yield nh.advertise('gazebo/set_model_state', ModelState)
+                try:
+                    x = float(args.gteleport.split(' ')[0])
+                    y = float(args.gteleport.split(' ')[1])
+                    z = float(args.gteleport.split(' ')[2])
+                except Exception:
+                    fprint("Format should be 'X Y Z' if you wish to input coords. Checking directory for saved location", msg_color='yellow')
```
kev-the-dev: don't really need this. It should only really print something on success or failure

command/sub8_missions/nodes/move_command
```diff
@@ -37,8 +47,49 @@ def main(args):
...
+                    try:
+                        argz = f[args.gteleport]
```
kev-the-dev: This will throw an vague key error exception if it isn't in the yaml. Instead you should always check that that key exsists
```
if args.teleport in f:
    # teleport
else:
   # notify user location not found
```

command/sub8_missions/nodes/tp.yaml
```diff
@@ -0,0 +1,6 @@
+bin: -24 14 -3.5
```
kev-the-dev: Add a comment in this file about what this data represents, what frame, units, and order of coordinates.

kev-the-dev (changes requested):
I couldn't get the teleportation to run on my system

command/sub8_missions/nodes/move_command
Grymessa: subc f 0 -gt buoy  no quotes required around buoy but I tested it and it works with quotes as well. I believe quotes are only required for coordinate inputs as it takes in a string now. For whatever reason every command seems to require that f 0 for me.  You can also do subc f 0 -gt X Y Z

command/sub8_missions/nodes/move_command
Grymessa: Wasn't quite sure how to do it manually so I did it this way. Still I was worried someone might not like it. Is there any reason this is particularly bad however?

command/sub8_missions/nodes/move_command
Grymessa: Ack! Sorry about that one.

command/sub8_missions/nodes/move_command
Grymessa: Guess it was more for clarification to the user but it can swamp the console a bit. I'll remove it.

command/sub8_missions/nodes/tp.yaml
Grymessa: Yeah since it was in that original warning that I removed I didn't think I needed to put it there. I'll do so now.

command/sub8_missions/nodes/move_command
Grymessa: I'll try that method out, see if I can get it to work. If I remember correctly I couldn't get it to work the first go around so I gave up and used try catches.  It also should throw the Saved location not found error message I wrote up.

command/sub8_missions/nodes/tp.yaml
DSsoto: This file should go under sub8_gazebo/config. It should also be called teleport_checkpoints or something like that rather than tp.

simulation/sub8_gazebo/README.md
```diff
@@ -18,21 +18,31 @@ And these can be combined if you'd like.
...
+This guide assumes you have created a mesh and texture within gazebo. All the models I have been given from Solidworks have already come with meshes so I did not have to create them. Look up guides on how to bind textures to UV Maps in Blender. That is what you will be doing for the most part to create your texture files. You'll want to save the texture file as a separate png for this. An interesting side note about Blender, the way the UV Maps work and meshes work, you cannot scale an object differently in each dimension and preserve the mesh. The entire object must be scaled at the same ratio (.5,.5,.5 for example). Otherwise when you export the object you will encounter many terrible errors trying to load it into Gazebo.
```
DSsoto, line 23: You should mention that you are currently maintaining this so people know who to contact.

command/sub8_missions/nodes/move_command
DSsoto: This should be implemented in the same way that the stop command was.
Usage would ideally be `subc teleport buoy` and if you add a shorthand we could do `subc tp buoy`

command/sub8_missions/nodes/move_command
```diff
@@ -37,8 +47,49 @@ def main(args):
         fprint("Waiting for odom...")
         yield sub.tx_pose()
         fprint("Odom found!", msg_color='green')
-
+        rospack = rospkg.RosPack()
+        config_file = os.path.join(rospack.get_path('sub8_missions'), 'nodes', 'tp.yaml')
```
DSsoto: Loading this config should only happen when `subc` is called with the teleport command.

command/sub8_missions/nodes/move_command
```diff
@@ -95,6 +146,8 @@ if __name__ == '__main__':
         help="Make end goal have zero roll and pitch.")
     parser.add_argument('-b', '--blind', action='store_true',
         help="Do not check if waypoint is safe.")
+    parser.add_argument('-gt', '--gteleport',metavar='object', type=str,
+        help="Teleport sub in Gazebo to object. Does nothing if not in Gazebo. Takes a string meaning quotes required around input if it is not a shortcut implemented through yaml.")
```
DSsoto: A couple of points although I don't think this should be implemented as an option:
* Metavars are usually short and capitalized
* Object isn't a very descriptive name. I would go for something like 'checkpoint'
* You usually want help messages to be a bit more concise

DSsoto (dismissed):
Requesting a couple of changes, especially implementing teleport as a command and not an option.

Grymessa:
Latest Amend:
Made it a command
Removed debug junk
Added some models (squid) while I was waiting on response - those don't really change any of the magic missile code but they are in sim now. 
Fixed Metavar I think and shortened help message
Changed name of file and moved it to correct config folder

command/sub8_missions/nodes/move_command
```diff
@@ -37,14 +47,53 @@ def main(args):
...
+        elif command in ['tp', 'teleport']:
+            try:
+                state_set_pub = yield nh.advertise('gazebo/set_model_state', ModelState)
+                config_file = os.path.join(rospack.get_path('sub8_gazebo'), 'config', 'teleport_locs.yaml')
+                f = yaml.load(open(config_file, 'r'))
+                try:
+                    x = float(argument.split(' ')[0])
+                    y = float(argument.split(' ')[1])
+                    z = float(argument.split(' ')[2])
+                except Exception:
```
DSsoto: what kind of exceptions are you hoping to catch here? It's ususally bad practice to catch all exception types in one block. Yo should do:
```python
except Exception1, Exception2:
     # Handle these specific exceptions
```

command/sub8_missions/nodes/move_command
```diff
@@ -37,14 +47,53 @@ def main(args):
         fprint("Waiting for odom...")
         yield sub.tx_pose()
         fprint("Odom found!", msg_color='green')
-
+        rospack = rospkg.RosPack()
```
DSsoto: This line should go under the teleport branch below. No need to even create a smaller name for something that's only used once though.

command/sub8_missions/nodes/move_command
```diff
@@ -95,6 +144,8 @@ if __name__ == '__main__':
         help="Make end goal have zero roll and pitch.")
     parser.add_argument('-b', '--blind', action='store_true',
         help="Do not check if waypoint is safe.")
+    parser.add_argument('-tp', '--teleport',metavar='Loc', type=str,
```
DSsoto: This option is no longer needed.

kev-the-dev (approved):

command/sub8_missions/nodes/move_command
sentree: True, we definitely want KeyboardInterrupt exceptions to terminate the program.

sentree:
Make sure that your formatting is correct by having a linter run automatically in your IDE @Grymestone. I think I remember seeing you running Sublime, so it should be fairly straightforward.

command/sub8_missions/nodes/move_command
Grymessa: So in my current iteration on my comp, I have added checks for the types of exceptions that will directly error out the commands being run, namely: IndexError, ValueError, and LookupError. At the end of this entire block of code though is a final catch that catches all Exception types, in the event we run into an unexpected error. Is that safe to have or should I just make that final catch a KeyboardInterrupt Exception and terminate there. 
Also as far as termination goes, I am using the "break" keyword. Is there something better I should use like sys.exit?

command/sub8_missions/nodes/move_command
```diff
@@ -5,8 +5,17 @@ from twisted.internet import defer, reactor
 import numpy as np
 from mil_misc_tools import text_effects
 from sub8 import sub_singleton
-from geometry_msgs.msg import PoseStamped, PointStamped
+from geometry_msgs.msg import PoseStamped, PointStamped, Pose, Point, Quaternion, Twist
+from gazebo_msgs.msg import LinkStates, ModelState
```
kingkevlar: It looks like LinkStates, publisher, Float64, BaggerCommands and HandlerBase all get imported but never used. 

In particular in BaggerCommands should be removed since @ironmig's recent changes to online bagger mean it is now unavailable.

command/sub8_missions/nodes/move_command
DSsoto: I think what you have is enough. You don't need to handle all possible exceptions for this type of thing. I think its fine to just let something like `KeyboardInterrupt` crash the program.

kev-the-dev:
@Grymestone what's up? We'd really like these sim changes. Address the few remaining comments and format that one file detected by CI:
```
sudo apt-get install clang-format-3.8
wget -O ~/.clang-format https://raw.githubusercontent.com/uf-mil/installer/master/.clang-format
clang-format-3.8 -style=file -i $CATKIN_DIR/src/Subjugator/simulation/sub8_gazebo/src/sub8_state_set.cpp
```

kev-the-dev:
I verified that everything David requested was fixed so I'm gonna merge this. I'm a little worried about the timing dependent stuff, but we can improve it later if it's a problem. Good work on this! SIM is an increasingly ghost-free environment to test in.

PR #284, opened 2017-06-23
BUOY: minor fixes to perception

kev-the-dev:
* Went through all bags I have of buoys to refine threshold values so all 3 buoys are correctly masked in each all bags
* Added circle-ness test before selecting a buoy contour, so random large noise contours should be excluded
* Made min contour area a proportion of image area so it is resistant to resolution changes.
* EDIT: update, integrated new Threshold and CircleFinder classes to replace some of the code


command/sub8_launch/config/buoy_finder.yaml
```diff
@@ -1,9 +1,10 @@
...
-min_observations : 8 # Observations needed to attempt a least squares position estimate
-min_contour_area : 10 # Minimum pixels for a pottential buoy contour to not be rejected
-max_velocity     : 1 # Highest velocity sub can be moving as in m/s before observations are ignored
-min_trans        : 0.05 # Smallest movement of sub in meters between observations to count towards observation count
-timeout_seconds  : 2.5 # Time in which min_observations must be seen to be considered found
+camera_topic      : /camera/front/right/image_rect_color #Image topic to use to find buoys
+min_observations  : 8 # Observations needed to attempt a least squares position estimate
+min_contour_area  : 0.002 # Minimum of image area contour must be to not be rejected
+max_velocity      : 1 # Highest velocity sub can be moving as in m/s before observations are ignored
+max_circle_error  : 0.4 #max 0-1 error in contour area vs enclosing circle area
```
DSsoto: How does this do when there are bright spots on the surface that remove parts of the contour?

perception/sub8_perception/nodes/buoy_finder.py
```diff
@@ -255,9 +259,10 @@ def image_cb(self, image):
         # Crop out some of the top and bottom to exclude the floor and surface reflections
         height = image.shape[0]
         roi_y = int(0.2 * height)
```
DSsoto: Did you choose these roi constants for operating at a particular depth?

perception/sub8_perception/nodes/buoy_finder.py
```diff
@@ -255,9 +259,10 @@ def image_cb(self, image):
...
-        roi_height = height - int(0.2 * height)
+        roi_height = height - int(0.1 * height)
         self.roi = (0, roi_y, roi_height, image.shape[1])
         self.last_image = image[self.roi[1]:self.roi[2], self.roi[0]:self.roi[3]]
+        self.image_area = self.last_image.shape[0] * self.last_image.shape[1]
```
DSsoto: Use `self.last_image.size`

perception/sub8_perception/nodes/buoy_finder.py
```diff
@@ -297,15 +302,18 @@ def get_best_contour(self, contours):
         TODO: Use smarter contour filtering methods, like checking this it is circle like
```
DSsoto: typo? if?

perception/sub8_perception/nodes/buoy_finder.py
```diff
@@ -297,15 +302,18 @@ def get_best_contour(self, contours):
...
-            area = cv2.contourArea(cnt)
-            if area < self.min_contour_area:
-                return None
-            M = cv2.moments(cnt)
-            cx = int(M['m10'] / M['m00'])
-            cy = int(M['m01'] / M['m00'])
-            tpl_center = (int(cx), int(cy))
-            return cnt, tpl_center, area
+            sort = sorted(contours, key=cv2.contourArea, reverse=True)
+            for cnt in sort:
```
DSsoto: Can you add a little header comment for this loop? Also, if your measure of contour changes, it might not be so good to have the return statement inside of the loop.

DSsoto (approved):
Minor comments

perception/sub8_perception/nodes/buoy_finder.py
kev-the-dev: Well that would be 3 times what I want because it is a 3 channel image.

DSsoto: Well, it can be used to get the right answer is all I'm saying. haha

command/sub8_launch/config/buoy_finder.yaml
kev-the-dev: It accepts a pretty big error, so it still works in these cases with the bags I tested with. It more just prevents the extreme case of a giant blob that looks nothing like a circle, like the walls of ground, being the largest thresholded contour.

But yes, it might need to be tuned for these cases.

perception/sub8_perception/nodes/buoy_finder.py
kev-the-dev: They're tuned to the bags I have to remove the worst noise in the bottom and top of frame. I changed them to ros params so they could be easily disabled or tuned if needed now.

kev-the-dev:
Addressed comments and replaced some code with stuff from mil_common. Just bumped in SIM again so no regressions. Should be more robust "circle" test using Hu invariants.

PR #285, opened 2017-06-23
COLLECT CLASSIFY: add rough draft mission code

kev-the-dev:
Add a rough draft of a mission to complete the entire collect classify mission (grab pipe, surface, place pipe on oval table). Parts of this is tested, but the rest will be once we have either working perception for these tasks or the models in SIM combined with @Grymestone's perception faker node.

Note: depends on some features in #286

### DIFF
--- a/command/sub8_missions/missions/collect_classify.py
+++ b/command/sub8_missions/missions/collect_classify.py
@@ -32,12 +32,12 @@ class CollectClassifyMission(object):
     A brief overview of the flow of the mission:
     * move in the direction of the pinger's heading until a pinger position estimate is available
     * Once pinger position is found, for each of the 4 tower pipers:
-     * Circle around pinget while facing tower looking for a pipe
+     * Circle around pinger while facing tower looking for a pipe
      * Once a pipe is found, grab pipe with gripper arm
      * backup and surface with pipe in gripper arm
      * submerge to depth above tower
      * If no ovals have been found:
-       * Circle around octogon (large radius around pinger) until an oval on the table is found
+       * Circle around octagon (large radius around pinger) until an oval on the table is found
      * Move above a previously found oval
      * Circle around oval table to find oval with same color as pipe in gripper
      * Once oval is found: move on top of oval and aligned with the longer axis
@@ -49,11 +49,12 @@ class CollectClassifyMission(object):
     BLIND = True
     BASE_LINK_TO_ARM = np.array([0, 0, 0])  # Transformation from base_link to claw of arm mechanism
     COLORS = ['red', 'green', 'orange', 'blue']
-    PIPE_SEARCH_HEIGHT = 0.6096  # Distance above ground to circle tower to find pipes, meters to dvl
+    PIPE_SEARCH_HEIGHT = 0.6096  # Distance above ground to circle tower to find pipes, meters to DVL
     PIPE_SEARCH_RADIUS = 2.0  # Radius in meters around pinger to circle tower to find pipes
-    OVAL_SEARCH_HEIGHT = 1.524  # Distance above ground to look for oval table, meters from ground to dvl
-    OVAL_DROP_HEIGHT = 1.0  # Height to drop pipe onto table, meters from ground to dvl
-    OCTOGON_RADIUS = 2.7  # Radius of octogon from rules, meters
+    OVAL_SEARCH_HEIGHT = 1.524  # Distance above ground to look for oval table, meters from ground to DVL
+    OVAL_DROP_HEIGHT = 1.0  # Height to drop pipe onto table, meters from ground to DVL
+    OCTOGON_RADIUS = 2.7  # Radius of octagon from rules, meters
+    SAFE_SPEED = 0.3  # Proportion of sub's max speed to execute moves where collisions are possible (grabbing pipe)
 
     def __init__(self, sub):
         self.sub = sub
@@ -67,12 +68,6 @@ class CollectClassifyMission(object):
 
         self.print_bad = FprintFactory(title=MISSION, msg_color="red").fprint
         self.print_good = FprintFactory(title=MISSION, msg_coolr="green").fprint
-
-    def height_to_depth(self, height):
-        '''
-        Uses dvl to convert a height over ground to a depth
-        '''
-        return -self.sub.move.to_height(height)._pose.position[2]
 
     @staticmethod
     def vision_response_to_pose(res):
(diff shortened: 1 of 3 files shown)


command/sub8_missions/missions/collect_classify.py
```diff
@@ -273,7 +280,8 @@ def find_oval(self, color):
         yield self.sub.vision_proxies.table_ovals.start()
 
         # Go to safe depth so we don't hit tower
-        yield self.sub.move.depth(CollectClassifyMission.OVAL_SEARCH_DEPTH).go()
+        depth = self.height_to_depth(CollectClassifyMission.OVAL_SEARCH_HEIGHT)
+        yield self.sub.move.depth(depth).go()
```
mattlangford: Why not just `sub.move.to_height(CollectClassifyMission.OVAL_SEARCH_HEIGHT)`?

command/sub8_missions/missions/collect_classify.py
```diff
@@ -320,11 +328,11 @@ def drop_pipe(self, pose):
         # Move at safe depth above oval
         position = pose[0]
         position = position + CollectClassifyMission.BASE_LINK_TO_ARM
-        position[2] = CollectClassifyMission.OVAL_SEARCH_DEPTH
+        position[2] = -self.height_to_depth(CollectClassifyMission.OVAL_SEARCH_HEIGHT)
         yield self.sub.move.set_position(position).set_orientation(pose[1]).zero_roll_and_pitch()
```
mattlangford: Similarly you could add `.to_height(...)` at the end of these as well. It would make it more obvious what the intent was, though it would make the line a lot longer.

command/sub8_missions/missions/collect_classify.py
```diff
@@ -0,0 +1,395 @@
...
+        yield self.sub.move.look_at(self.pinger_position).set_position(
+            self.pinger_position).backward(3.0).go(blind=self.BLIND)
+
+    @util.cancellableInlineCallbacks
+    def surface(self):
+        '''
+        Surface sub at current position, wait a few seconds,
+        then submerge again.
+        '''
+        yield self.sub.move.depth(0).go()
```
mattlangford: Will the move command let you move to a depth of 0?

command/sub8_missions/missions/collect_classify.py
```diff
@@ -0,0 +1,395 @@
...
+        yield self.sub.nh.sleep(2.0).go()
+        depth = self.height_to_depth(CollectClassifyMission.OVAL_SEARCH_HEIGHT)
+        yield self.sub.move.depth(depth).go()
+
+    @util.cancellableInlineCallbacks
+    def circle(self, origin, radius, num_moves=10, proportion=1.0):
```
mattlangford: I think they best place for this could be in the `_PoseProxy` class in the sub_singleton file. There are some "advanced moves" in there that this could fall under.

command/sub8_missions/missions/collect_classify.py
```diff
@@ -0,0 +1,395 @@
...
+        move = self.sub.move.set_position(pose[0]).set_orientation(pose[1]).yaw_left(
+            np.pi / 2.0).zero_roll_and_pitch()
+        # Adjust for gripper tf
+        move = move.relative(CollectClassifyMission.BASE_LINK_TO_ARM)
+        # Rotate 180 if we're not facing pinger
+        m = trns.quaternion_matrix(self.sub.move._pose.orientation).dot(np.array([0.1, 0, 0]))
+        if np.linalg.norm(m - self.pinger_position) > np.linalg.norm(move._pose.position - self.pinger_position):
+            move = move.yaw_left(np.pi)
+
+        move = move.backward(1.5)
```
mattlangford, line 174: I think a lot of these moves should be done with a non-default speed. Things get scary when the sub is moving as fast as it can. What do you think?

mattlangford (commented):
I didn't go through the logic of the mission in super detail, someone with a better understanding of the rules should do that.
I think the main question I have is why convert height to depth before moving and not just `to_height`?

command/sub8_missions/missions/collect_classify.py
kev-the-dev: Not currently. I guess we need some sort of override flag for this case.

kev-the-dev:
I guess I wanted to abstract away the get_height thing so I could throw in something else later because I didn't like that it measured (I think) from the DVL instead of base_link. But yeah I should just fix it in sub_singleton for everyone else.

command/sub8_missions/missions/collect_classify.py
kev-the-dev: I added a separate speed constant for sketchy moves like this.

command/sub8_missions/missions/collect_classify.py
mattlangford: You could use the `blind` keyword and modify [this line](https://github.com/uf-mil/SubjuGator/blob/master/command/sub8_missions/sub8/sub_singleton.py#L152) so that it'll be skipped with that.

command/sub8_missions/sub8/sub_singleton.py
```diff
@@ -144,12 +172,18 @@ def check_goal(self):
...
+        Set the sub's waypoint to _pose
+        @param check_feasibility Check that goal is 'feasibile' and correct otherwise, see check_goal
+        @return Defer object for action client to complete this move. yield on this to wait for move to complete
+        '''
         if self.print_only:
             print self._pose
             return self._sub.nh.sleep(0.1)
 
-        self.check_goal()
+        if check_feasibility:
```
mattlangford, line 187: I think this is pretty much the same as `blind` isn't it? I think that'd be a good keyword to piggyback off of

command/sub8_missions/missions/collect_classify.py
```diff
@@ -0,0 +1,387 @@
...
+    Class to carry out Collect and Classify mission in RoboSub2017.
+
+    Assumes perception services are available to report found pipes in the tower
+    and ovals on the table as 3D position and orientations and a perception service
+    is available to report 3D position of the pinger at the center of the tower.
+
+    A brief overview of the flow of the mission:
+    * move in the direction of the pinger's heading until a pinger position estimate is available
+    * Once pinger position is found, for each of the 4 tower pipers:
+     * Circle around pinget while facing tower looking for a pipe
```
DSsoto: pinger

command/sub8_missions/missions/collect_classify.py
```diff
@@ -0,0 +1,387 @@
...
+     * Once a pipe is found, grab pipe with gripper arm
+     * backup and surface with pipe in gripper arm
+     * submerge to depth above tower
+     * If no ovals have been found:
+       * Circle around octogon (large radius around pinger) until an oval on the table is found
```
DSsoto: octagon

command/sub8_missions/missions/collect_classify.py
```diff
@@ -0,0 +1,387 @@
...
+        Circle at octogon radius around pinger to find the oval table.
+        If an oval has already been found or if one is found during search pattern,
+        return True. Otherwise, return False.
+        '''
+        if len(self.ovals_found) > 0:
+            defer.returnValue(True)
+
+        origin = self.pinger_position
+        origin[2] = CollectClassifyMission.OVAL_SEARCH_DEPTH
+        circle_done = False
```
DSsoto: Did you mean `self.circle_done`? And in all the other places you use this.

command/sub8_missions/missions/collect_classify.py
```diff
@@ -0,0 +1,387 @@
...
+        search = circle()
+        while not circle_done:
+            res = yield self.get_oval()
+            if res is not None:
+                search.cancel()
+                defer.returnValue(True)
+        defer.returnValue(False)
+
+    @util.cancellableInlineCallbacks
+    def find_oval(self, color):
```
DSsoto, line 245: Should we decouple the finding of the table from finding the ovals and identifying color?
Using ColoredRectangleFinder?

DSsoto (commented):
I'd like to see this running in the sim before approving.

command/sub8_missions/missions/collect_classify.py
kev-the-dev: That wouldn't be a bad idea for the oval perception to implement, but I think at the mission level I shouldn't have to worry about it.

sentree: This would probably lead to better results, I will work on this once the oval perception is complete.

kev-the-dev:
Archived at [archive_pr_285](https://github.com/uf-mil/SubjuGator/tree/archive_pr_285)

PR #287, opened 2017-06-23
VIS, ALIAS: improve default rviz config, add subfmt alias

kev-the-dev:
* Add convenient ```subfmt``` alias which will check the pep8 validity of your Subjugator repo, so developers can check before submitting a PR
* Update sub.rviz with renamed topics (ex: "Trajectory" instead of generic "Arrow" topic), unchecked camera and blueview topics, and a more gray background color that stands out better against the black sub model


kev-the-dev:
rviz

DSsoto (approved):

PR #291, opened 2017-06-27
SUBC: add list option to teleport command

kev-the-dev:
* also make sure to yield on txros ros_alarms functions, so we know they finish before executing next line

### DIFF
--- a/command/sub8_missions/nodes/move_command
+++ b/command/sub8_missions/nodes/move_command
@@ -78,6 +78,11 @@ def main(args):
                         break
                     except ValueError:
                         try:
+                            if argument in ('list', 'l'):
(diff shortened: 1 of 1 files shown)


command/sub8_missions/nodes/move_command
```diff
@@ -78,6 +78,11 @@ def main(args):
                         break
                     except ValueError:
                         try:
+                            if argument in ('list', 'l'):
+                                fprint('Available TP locations:')
+                                for k in f:
```
DSsoto, line 83: Let's try to not name variables one letter.

DSsoto (approved):
We should probably start refactoring this into different functions for different commands.

PR #292, opened 2017-06-28
DIAGNOSTICS: add python SubjuGatorException base

DSsoto:
I think this will be convenient for catching exceptions that are very specific to our vehicle. I have especially found myself missing something like this while working on driver related code.


kev-the-dev (approved):
Good idea

PR #296, opened 2017-07-01
GNC: publish zero wrench when killed

kev-the-dev:
* ensures thruster feedback is generated even when killed


DSsoto (approved):
Good work, we'll have feedback all the time now.

PR #302, opened 2017-07-14
POSE SEQUENCE: Change commander to calculate out path in beginning

LucasBA:
This is an update to the pose sequence commander that calculates all the way points before moving rather than after each way point. I am working on updating the command line interface also but it is taking a while and I might need to totally redo it.


RustyBamboo:
@LucasBA What's the status of this?

kev-the-dev:
Archived at [archive_pr_302](https://github.com/uf-mil/SubjuGator/tree/archive_pr_302)

PR #307, opened 2017-07-18
SIM: Add lights, Add tower model, restructure course

Grymessa:
Added lights to fix the lighting issues, added what I think are the last of the models for the competition and ordered the simulator so that it follows the course of the actual competition. Everything should be in there.


RustyBamboo (approved):
Tubular isn't the most intuitive naming. Maybe something like tower_tubes?

simulation/sub8_gazebo/launch/duck.launch
```diff
@@ -25,53 +25,53 @@
...
-      -x -22.3 -y 14.16 -z -4.47 -Z -1.57 -model lid"
+      -x 31.7 -y 3.0 -z -4.47 -Z -1.57 -model lid"
       output="screen" />
 
   <!-- Attempting to spawn the start gate -->
   <node
       name ="start_gate" pkg="gazebo_ros" type="spawn_model"
       args="-file $(find sub8_gazebo)/models/start_gate/start_gate.sdf -sdf
-      -x 15 -y 10 -z -2 -Z -1.57 -model start_gate"
+      -x 3.22 -y 26.58 -z -2 -Z -1.57 -model start_gate"
```
kev-the-dev: This new location doesn't give much room to see the gate if the sub starts behind it.

Also, can you move the first path marker to be in front of the start gate and point towards buoys? This is how the course at the transdec will be.

Lastly, can you change the sub spawn to be behind the start gate by default (note this is hardcoded in a few places so be careful)

kev-the-dev (changes requested):
Thanks for the progress. Got a few more things before the course can be used to test chained missions.

Looks like this breaks the teleport option in move_command. Please either hard code in the new numbers or (better) refactor the command so it teleports based on the state of the gazebo model (like fake perception does). If I designed it, I would structure the yaml like this:
```
# For each teleport location
torpedoboard:
    model: squilliamson
    offset: -2.0 0.0 1.0 # Offset from model position to teleport sub to, x y z
...
```

simulation/sub8_gazebo/launch/duck.launch
```diff
@@ -25,53 +25,53 @@
...
-        -x 25 -y -15 -z -1.5 -Z -1.57 -model nav_gate"
+        -x 24 -y 16.15 -z -1.5 -Z -1.57 -model nav_gate"
         output="screen" />
 
     <!-- Attempting to spawn the octogon -->
     <node
         name ="octagon" pkg="gazebo_ros" type="spawn_model"
         args="-file $(find sub8_gazebo)/models/octogon/octogon.sdf -sdf
-        -x 20 -y -20 -z 0 -Z -1.57 -model octogon"
+        -x 38.8 -y 2 -z 0 -Z -1.57 -model octogon"
```
kev-the-dev: Also noticed that the beach balls need to be moved too, as they're just floating in no where now

kev-the-dev (changes requested):

RustyBamboo:
Branch was updated to different code

PR #309, opened 2017-07-22
CI: verify roslaunch files syntax and dependencies

kev-the-dev:
thought it would be a good idea to verify roslaunch files with CI, to be sure we never reference executables that don't exist as we've done on occasion


DSsoto (approved):
Good idea. It looks good to me. @whispercoros any comments?

kev-the-dev:
Doen in #383

PR #310, opened 2017-07-24
Passive sonar

DSsoto:
* Fixed the tf for hydrophones
* The tf publisher is necessary for using the simulated `ReceiverArrayInterface` class in mil_passive_sonar


DSsoto:
Superceded by #313

PR #317, opened 2017-07-28
GNC: Update thruster model to 4th order polynomial

DSsoto:
* Also tune gains for better performance w/ new model


RustyBamboo (approved):
Oh gez

jnez71:
4th order? woah baby
For future, we should move back to linearly-interpolated look-up tables.

DSsoto:
@jnez71 If you look at the data we had from video ray with their props and shrouds, this is essentially exactly that but with better interpolation.

PR #326, opened 2018-04-27
3DSim

RustyBamboo:
Work in progress. Does work to some extent, however ignores fixed_frame to base_link


kev-the-dev:
replaced by #382

PR #336, opened 2018-06-05
CAMERAS: add calibration and fix naming

RustyBamboo:


Grymessa (approved):
Getting rid of old calibration files, replacing them with new ones. Necessary due to switching camera mounts and moving to Seecam from pointgrey.

PR #340, opened 2018-06-11
BLUEVIEW: add more params, tune thresholds, change tf

RustyBamboo:
- TF for blueview removes the 5-deg tilt
- Add debugging and ogrid params
- Add threshold values for closeness and depth


Grymessa (approved):
Some simple checks, extra thresholds etc from the looks of it. Or at least reworked some logic on em looks like. Approved.

PR #341, opened 2018-06-11
STARTGATE: mission to search for two poles

RustyBamboo:
Work in progress.

### DIFF
--- a/command/sub8_missions/missions/start_gate.py
+++ b/command/sub8_missions/missions/start_gate.py
@@ -2,68 +2,93 @@
 from __future__ import division
 
 import txros
-import txros.tf as tf
 from twisted.internet import defer
 
 from mil_misc_tools import text_effects
 
-from sub8 import Searcher
+from sub8 import SonarObjects
 from mil_ros_tools import rosmsg_to_numpy
+from scipy.spatial import distance
 
 import numpy as np
 
 fprint = text_effects.FprintFactory(
     title="START_GATE", msg_color="cyan").fprint
 
-# Distance before and after the gate in meters
-FACTOR_DISTANCE_BEFORE = 1.5
-FACTOR_DISTANCE_AFTER = 1.5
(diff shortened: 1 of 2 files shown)


command/sub8_missions/missions/start_gate.py
```diff
@@ -2,68 +2,93 @@
...
-    point_after = position - FACTOR_DISTANCE_AFTER * normal
-
-    # go in front of gate
-    fprint('Moving infront of gate {}'.format(point_before))
-    yield sub.move.set_position(point_before).look_at(point_after).zero_roll_and_pitch().go(speed=SPEED)
-
-    # go through the gate
-    fprint('YOLO! Going through gate {}'.format(point_after))
-    yield sub.move.set_position(point_after).zero_roll_and_pitch().go(speed=SPEED)
+    fprint('Begin search for gates')
```
Grymessa, line 25: You should put some comments in here as to what this is doing and why it is necessary. Its not exactly obvious.

command/sub8_missions/missions/start_gate.py
```diff
@@ -2,68 +2,93 @@
...
+    rotate_start = sub.move.zero_roll_and_pitch()
+    for i in range(4):
+        fprint('Searching {} degrees'.format(90 * i))
+        yield rotate_start.yaw_right_deg(90 * i).go(speed=SPEED)
+        start = sub.move.zero_roll_and_pitch()
+        so = SonarObjects(sub, [start.pitch_down_deg(10), start])
```
Grymessa: Might want to explain this a bit too but if I recall you added comments in a different mission so maybe just direct people to that in a one line comment.

Grymessa (approved):
Hesitant approval since this has been proven to work more or less but it does overshoot the start gate and could use a few more comments. But those are minor gripes and working code is more important right now.

PR #345, opened 2018-06-14
PATH_MARKER: Add trained model for path detection and utils

pablovela5620:


Grymessa:
I'm glad you got this working man nice job. A few issues here though is there are 70 files in this pull request. I know you only need like three or four of them to actually run the mission. If you wanted to add all the utilities you used to train the model, I would suggest separating those for now and isolating the ones that are strictly necessary to do so. I noticed a lot of meta files and checkpoints that as far as I can tell are useless. Find the files you need and remove the others from the PR and I can get to looking at it more deeply and giving some better feedback or just merging it if it doesn't need anything else.

PR #346, opened 2018-06-15
DICE: just use camera ray for now

RustyBamboo:


command/sub8_missions/missions/dice.py
```diff
@@ -66,30 +68,9 @@ def run(sub):
         # Get one of the dice
         dice, xy = found.popitem()
         fprint('dice {}'.format(dice))
-        ray, base = yield get_transform(sub, model, xy)
-
-        so = SonarObjects(sub, [
```
jpanikulam, line 71: I dreamed a dream in time gone by

RustyBamboo: oof

PR #350, opened 2018-07-07
AUTONOMOUS: add higher level timeouts

RustyBamboo:


command/sub8_missions/missions/autonomous.py
```diff
@@ -2,22 +2,35 @@
...
+
+
 @txros.util.cancellableInlineCallbacks
 def do_mission(sub):
     fprint("RUNNING MISSION", msg_color="blue")
 
     # Chain 1 missions
     try:
-        yield start_gate_simple.run(sub)
+        yield run_mission(sub, start_gate, 300)
```
Grymessa, line 33: This should be in a config file

Grymessa (approved):
Looks good, I would suggest using a config file similar to the mission for torpedo board so you dont have to keep pushing one line updates if you want to change the time out.

PR #352, opened 2018-07-09
CAMERAS: fix down cam and add relaunch on error

RustyBamboo:


Grymessa (approved):
Tested and confirmed it works. You are granted the Stone Cold Seal of Approval hereafter referred to as SCSA.

PR #353, opened 2018-07-12
POOL: address and fix misc

RustyBamboo:
* rise coefs
* down cam frame rate
* depth bug in active sonar
* increase Heartbeat monitor duration to avoid unintended autonomous misfiring

### DIFF
--- a/command/sub8_alarm/alarm_handlers/network_loss.py
+++ b/command/sub8_alarm/alarm_handlers/network_loss.py
@@ -6,7 +6,7 @@ class NetworkLoss(HandlerBase):
     alarm_name = 'network-loss'
 
     def __init__(self):
-        self.hm = HeartbeatMonitor(self.alarm_name, "/network", Header, node_name="network_loss_kill")
+        self.hm = HeartbeatMonitor(self.alarm_name, "/network", Header, prd=0.8, node_name="network_loss_kill")
 
     def raised(self, alarm):
         pass
--- a/command/sub8_launch/config/rise_6dof.yaml
+++ b/command/sub8_launch/config/rise_6dof.yaml
@@ -1,8 +1,8 @@
-accel_feedforward: 0, 0, 0, 0, 0, 0
+accel_feedforward: 0, 0, 0, 0, 0, 1
 alpha: 0.3, 0.3, 0.3, 0.3, 0.3, 0.3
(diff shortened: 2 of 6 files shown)


command/sub8_alarm/alarm_handlers/network_loss.py
```diff
@@ -6,7 +6,7 @@ class NetworkLoss(HandlerBase):
     alarm_name = 'network-loss'
 
     def __init__(self):
-        self.hm = HeartbeatMonitor(self.alarm_name, "/network", Header, node_name="network_loss_kill")
+        self.hm = HeartbeatMonitor(self.alarm_name, "/network", Header, prd=0.8, node_name="network_loss_kill")
```
Grymessa, line 9: What does prd do?

command/sub8_launch/config/rise_6dof.yaml
```diff
@@ -1,8 +1,8 @@
-accel_feedforward: 0, 0, 0, 0, 0, 0
+accel_feedforward: 0, 0, 0, 0, 0, 1
```
Grymessa, line 1: We should get a reference sheet going for the gains. Not sure how important it will be if we do move to sub9 but on the very high chance we don't make the move we will need it.

perception/sub8_pointcloud/src/OGridGen.cpp
```diff
@@ -159,11 +159,8 @@ void OGridGen::callback(const mil_blueview_driver::BlueViewPingPtr &ping_msg)
       point.x = newVec.x() + transform_.getOrigin().x();
       point.y = newVec.y() + transform_.getOrigin().y();
       point.z = newVec.z() + transform_.getOrigin().z();
-      // Ignore points if they are below the ground. This will depend on the current return from the depth.
-      if (point.z + dvl_range_ > 0)
-        continue;
       // Ignore points if they are below some depth in map frame
-      else if (point.z < -params.depth)
+      if (point.z < -params.depth)
```
Grymessa, line 163: Reasoning behind moving away from dvl range? I imagine it has to do with the slope of the pool and so on.

Grymessa (changes requested):
Add a quick comment to explain what prd is/what it does.

Grymessa:
Approved.

PR #357, opened 2018-07-18
CONTROLLER: use all terms not PD

RustyBamboo:


legacy/rise_6dof/src/rise_6dof/controller.py
```diff
@@ -60,10 +66,12 @@ def update(self, dt, desired, current):
 
         error_velocity_world = (desired_x_dot + body_gain(
             numpy.diag(self.config['k'])).dot(error_position_world)) - x_dot
```
Grymessa, line 68: So k is error in position

legacy/rise_6dof/src/rise_6dof/controller.py
```diff
@@ -60,10 +66,12 @@ def update(self, dt, desired, current):
...
+        print('{:20s} {:30}'.format('Error velocity:', error_velocity_world))
```
Grymessa, line 69: Ks is velocity error, this is starting to make sense. I like the print statements btw, very nice.

legacy/rise_6dof/src/rise_6dof/controller.py
```diff
@@ -60,10 +66,12 @@ def update(self, dt, desired, current):
...
         if self.config['two_d_mode']:
```
Grymessa, line 70: I love the effort they went through to maintain proper syntax. 'two_d_mode' lol

legacy/rise_6dof/src/rise_6dof/controller.py
```diff
@@ -5,6 +5,8 @@
 from tf import transformations
 from sub8.pose_editor import quat_to_rotvec
 
+numpy.set_printoptions(suppress=True, linewidth=130)
```
Grymessa, line 8: Might be worth adding this to config my dude.

command/sub8_launch/config/rise_6dof.yaml
```diff
@@ -1,8 +1,14 @@
-accel_feedforward: 0, 0, 0, 0, 0, 1
-alpha: 0.3, 0.3, 0.3, 0.3, 0.3, 0.3
-beta: 0, 5, 0, 0, 0, 0
-k: 3, 3, 3.0, 1.3, 1.3, 2.0
-ks: 114, 114, 171, 20, 50, 50
-two_d_mode: false
+# RISE terms:
```
Grymessa: You know, instead of zeroing these out why not just make use_rise false? Seems like that was the intended use of that flag lol.

Grymessa (approved):
Very useful changes for understanding how this stuff works and hopefully we get a more stable sub out of your changes to the gains. I would suggest adding the print statements to the config for ease of use.

PR #358, opened 2018-07-19
SIM: Make physics an option

Grymessa:


RustyBamboo:
You should be doing `git pull --rebase upstream master` ... instead of the `git pull` or `git merge` when updating your branch

simulation/sub8_gazebo/nodes/gazebo_controller.py
```diff
@@ -56,7 +58,7 @@ def send_wrench(self, wrench):
             body_name=self.target,
             reference_frame='world',
             wrench=wrench,
-            duration=rospy.Duration(0.01)
+            duration=rospy.Duration(0.001)
```
RustyBamboo, line 61: was the intentional?

Grymessa: Not exactly but it didn't seem to change performance. Can change it back if desired.

RustyBamboo (approved):

PR #360, opened 2018-07-21
Add Guess Points

LucasBA:
I added the service to request a point by its name. The naming is done by filling strings in correct order into the items list. I did this so that the user could type in all the item names before the sub is in the water during a competition run (somewhat time intensive and typo risky). Then the user selects points in rviz once the sub is in the water.


LucasBA:
I am not really sure how I could have broken velodyne_driver is this a known issue?

PR #367, opened 2018-09-25
sub8_hydrophones: fix to 9b5c452f14491dccbbf6cafba65856efc463d68b which renamed package but didn't fix scripts

forrestv:


RustyBamboo:
Keep the filenames with `.py`?

forrestv:
Why? They're executables, not modules. The .py being removed actually
broke something: renaming hydrophones to hydrophones.py made it so
`import hydrophones` in the other scripts ran the hydrophones script
rather than importing the hydrophones library.

forrestv:
Err, *adding* the .py broke things.

PR #368, opened 2018-10-12
Hydrophones fixes

forrestv:
* allow running hydrophones script directly on bag for offline analysis/debugging
* don't bother trying to find start position of ping, instead just trust low-level driver's triggering
* use pearson correlation instead of mean absolute difference to find relative delays, which should be more robust
* intelligently compute search window based on hydrophone spacing rather than assuming they're within 1 wavelength of each other
* optionally generate plots for debugging

See https://pastebin.com/fUn2Um0d for the results of running on the bags at https://ci.mil.ufl.edu/labelme/bags/navigator/2018-10-07/; results look decent, except for one outlier (marked) that is due to the motors being turned back on during the recording.

### DIFF
--- a/drivers/sub8_hydrophones/scripts/hydrophones
+++ b/drivers/sub8_hydrophones/scripts/hydrophones
@@ -2,14 +2,15 @@
 
 import numpy
 import math
+import sys
 
+import rosbag
 import rospy
 
 from sub8_hydrophones import algorithms, util
 from sub8_hydrophones.msg import Ping, ProcessedPing, Debug
 from std_msgs.msg import Header
 from geometry_msgs.msg import Point, PoseStamped, Pose
-
 
 def process_ping(ping):
     ping.header.frame_id = 'hydrophones'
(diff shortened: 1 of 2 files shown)


drivers/sub8_hydrophones/scripts/hydrophones
```diff
@@ -58,5 +61,11 @@ template_periods = rospy.get_param('~template_periods', 3)
...
 pub_debug = rospy.Publisher('hydrophones/debug', Debug, queue_size=10)
-sub = rospy.Subscriber('hydrophones/ping', Ping, process_ping)
-rospy.spin()
+
+if rospy.myargv()[1:]:
+    bag = rosbag.Bag(rospy.myargv()[1])
+    for msg in bag.read_messages(['/hydrophones/ping']):
+        process_ping(msg.message)
+else:
+    sub = rospy.Subscriber('hydrophones/ping', Ping)
```
kev-the-dev: Looks like this accidentally prevented the callback from being called, fixed in e39b9336d772

PR #403, opened 2019-06-27
Fix dvl pose in urdf

kev-the-dev:
So I had to put in a hack to get sim working and forgot to fix it. I placed the dvl lower so the ray tracing didn't collide with the CAD model.... unfortunately this meant the static transform would be wrong to. I guess this wouldn't cause many problems but it might.

I put the pose back to the right one (see rviz to verify) and added an attribute to the gazebo plugin to move it in gazebo a little lower so it simulates correctly.

REQUIRES uf-mil/mil_common#202


RustyBamboo:
Papa bless

PR #404, opened 2019-06-28
MISSIONS: startgate, pinger, and vampire

RustyBamboo:
Startgate:
    - add fast and slow speed
    - do full rotation

Pinger:
    - Correct transformations by using sub current position
    - Typo in tolerance

Vampire:
    - The beginning of greatness


command/sub8_missions/sub8_missions/autonomous.py
```diff
@@ -69,8 +70,9 @@ def do_mission(self):
                         yield self.move.yaw_right_deg(179).go(speed=0.4)
                         yield self.move.yaw_right_deg(179).go(speed=0.4)
 
+            fprint("Vampire Slayer")
```
kev-the-dev, line 73: `self.send_feedback` has the advantage of printing these user feedback messages in the client terminal that ran the mission or in the mission GUI, which is probably more useful

perception/sub8_perception/nodes/guess_server.py
```diff
@@ -11,7 +11,7 @@ class Guess:
     def __init__(self):
```
kev-the-dev, line 11: Isn't this node basically just `mil_poi`? Nice to avoid duplicate utilities

kev-the-dev (commented):
Sweet to see so much progress on missions

PR #405, opened 2019-06-30
Simulate pinger

kev-the-dev:
Requires uf-mil/mil_common#203

* Add a simulated pinger using the mil_common plugin. This does not actually simulate acoustics, but just sends out a vector from the reference sensor to a model in gazebo.
* Add passive_sonar into the main launch, so it runs when the sub is launched. This will also be sure to show the pinger visualization in the sim. I'm not sure if there was a particular reason this was not already the cast, so lmk if there was reason
* Modify the default rviz config to show the pinger heading visualziation

This should make it possible to simulate the pinger mission in the sim :)

### DIFF
--- a/command/sub8_launch/launch/subsystems/passive_sonar.launch
+++ b/command/sub8_launch/launch/subsystems/passive_sonar.launch
@@ -1,10 +1,14 @@
+<?xml version="1.0" encoding="UTF-8" ?>
 <launch>
-  <node pkg="sub8_launch" type="passive_sonar_conn" name="passive_sonar_conn" />
-  <node pkg="mil_passive_sonar" type="hydrophones" name="hydrophones_heading">
-    <param name="dist_h" value="2.286e-02" />
-    <param name="dist_h4" value="2.286e-02" />
-    <param name="v_sound" value="1482" />
-  </node>
+  <arg name="environment" default="real" />
+  <group if="$(eval environment == 'real')">
+    <node pkg="sub8_launch" type="passive_sonar_conn" name="passive_sonar_conn" />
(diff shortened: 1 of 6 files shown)


simulation/sub8_gazebo/worlds/a_whole_new.world
```diff
@@ -130,6 +130,12 @@
       </link>
     </model>
 
+    <include>
+      <name>transdec_pinger</name>
+      <uri>model://pinger</uri>
+      <pose>15 0 -2 0 0 0 </pose>
```
kev-the-dev, line 136: Totally random starting position here, please change to something more appropriate

command/sub8_launch/launch/subsystems/passive_sonar.launch
```diff
@@ -1,10 +1,14 @@
...
-  <node pkg="mil_passive_sonar" type="hydrophones" name="hydrophones_heading">
-    <param name="dist_h" value="2.286e-02" />
-    <param name="dist_h4" value="2.286e-02" />
-    <param name="v_sound" value="1482" />
-  </node>
+  <arg name="environment" default="real" />
+  <group if="$(eval environment == 'real')">
+    <node pkg="sub8_launch" type="passive_sonar_conn" name="passive_sonar_conn" />
+    <node pkg="mil_passive_sonar" type="hydrophones" name="hydrophones_heading">
+      <param name="dist_h" value="2.286e-02" />
```
kev-the-dev, line 7: Note that this change will launch the passive_sonar_conn and hydrophones nodes with the main launch. Be sure to test on the sub before merging

kev-the-dev (commented):
Some cliff notes for the skimming reviewer

PR #409, opened 2019-07-23
add gazebo movement test

DavidSadowsky:


utils/sub8_diagnostics/src/move_test.cpp
```diff
@@ -0,0 +1,72 @@
...
+
+    // checks current robot state vs. goal state
+    if(ac.getState() == actionlib::SimpleClientGoalState::SUCCEEDED){
+    cout << "The sub moved " + x + " meter(s) forward" << endl;
+    if(argc == 3) {
+      cout << "The sub moved " + y + " meter(s) up";
+    }
+  }
+    else {
+    cout << "The sub failed to move forward " + x + " meter(s)" << endl;
```
kev-the-dev, line 66: Indention is all over here. Try `clan-format -i <file`> to autoformat your code to our standards.

utils/sub8_diagnostics/CMakeLists.txt
```diff
@@ -1,18 +1,67 @@
...
+  mil_msgs
+  actionlib
+  message_generation
+  geometry_msgs
+  sensor_msgs
+  interactive_markers
+  std_msgs
+  actionlib_msgs
+  mil_tools
+)
```
kev-the-dev, line 20: Only some of these dependencies are actually needed (check the includes in your code). While adding more doesn't change program behavior, it does complicate the build process and make it more confusing to read

utils/sub8_diagnostics/src/move_test.cpp
```diff
@@ -0,0 +1,72 @@
...
+    // declares goal for the sub
+    mil_msgs::MoveToGoal goal;
+    x = string(argv[1]);
+
+    goal.header.frame_id = "base_link";
+    goal.header.stamp = ros::Time::now();
+
+    // sets goal to x meter forward
+    goal.posetwist.pose.position.x = atoi(argv[1]);
+    goal.posetwist.pose.orientation.w = atoi(argv[1]);
```
kev-the-dev, line 44: Why is orientation.w being set to the same as position.x? I'm having trouble seeing the usecase for that

kev-the-dev (commented):
Nice progress on this. Not sure if @RustyBamboo had other plans for this test than I originally envisioned, but I have some feedback for the design.

Basically, does this do anything that `submove f <x> right <y>` doesn't? It seems to just send a move and see if the motion planner thinks it got there, which is what `submove` does

I'd suggest a few things:
* Verify with `/odom` if it actually got there (don't trust the motion planner, as this is the thing you are testing)
* Turn this into a rostest so it could be run as part of the ci process, which would be super useful!
* Hard code some common moves if user doesn't specify them

PR #411, opened 2019-07-29
Mil poi integration

MarshallRawson:
removes any calls to guess server. does not change anything in deprecated and leave guess server currently. This may need to be changed? please give feed back

### DIFF
--- a/command/sub8_missions/sub8_missions/ball_drop.py
+++ b/command/sub8_missions/sub8_missions/ball_drop.py
@@ -13,6 +13,7 @@ from sub8_msgs.srv import GuessRequest, GuessRequestRequest
 
 import mil_ros_tools
 import rospy
+from std_srvs.srv import Trigger
 
 fprint = text_effects.FprintFactory(title="BALL_DROP", msg_color="cyan").fprint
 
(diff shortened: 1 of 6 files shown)


RustyBamboo:
Good stuff

command/sub8_missions/sub8_missions/ball_drop.py
```diff
@@ -60,17 +61,17 @@ def run(self, args):
      #   yield enable_service(SetBoolRequest(data=True))
 
         try:
-            vamp_txros = yield self.nh.get_service_client('/guess_location',
-                                                          GuessRequest)
-            ball_drop_req = yield vamp_txros(GuessRequestRequest(item='ball_drop'))
-            if ball_drop_req.found is False:
+            save_pois = rospy.ServiceProxy(
```
kev-the-dev, line 64: Using `rospy` and `txros` at the same time is going to break things.

This is why I made the `TxPoiClient ` [used in navigator](https://github.com/uf-mil/NaviGator/blob/master/mission_control/navigator_missions/navigator_missions/navigator.py#L113)

kev-the-dev (commented):
oof, as the kids say

